PR #21, opened 2020-01-21
Make changes to solve the problems with AOT compilation and deploy

ps37:
examples app to Firebase

**Why was this change made?**
To solve the following problems with AOT compilation:

Problem:
-Function calls inside the module metadata in DocLib module's static
forRoot function is causing following error which is not useful for
debugging:
ERROR: Function calls are not supported in decorators.
https://www.bennadel.com/blog/3565-providing-module-configuration-using-forroot-and-ahead-of-time-compiling-in-angular-7-2-0.htm

Workaround:
-Use a factory function declared outside the ngModule and call the
constructor 'new' function of doc retriever service inside that. And use
a intermediate injection token to provide the deps required for creation of the factory object.

Problem:
-Importing from barrels within a library is causing the following error:
ERROR in Unexpected value 'undefined' imported by the module
https://github.com/angular/angular/issues/23713

Workaround:
-Within the library packages, I am not importing files from barrels.
Instead they are being imported from absolute paths.

Problem:
-Loading of routes in the existing way is causing following error:
ERROR in Cannot read property 'loadChildren' of undefined
https://stackoverflow.com/questions/44233195/dynamically-adding-routes-in-angular

Workaround:
-Use resetConfig in appComponent to reset the routes instead of setting them in app routing module

Problem: This happens in the web browser
-Documentation JSONS being provided from app component are becoming null inside the doc lib module utilizing them

Workaround:
-Assign the doc jsons imported to exported variables before calling the module static forRoot function with them

**Testing Done:**
-Deployed the app to a Firebase end point and it's now [live here](https://vcdcc-examples-app.firebaseapp.com)
-Verified that the bundle size is smaller than the one being downloaded
during development
-Verified that the app at the end point works as expected

**Bug Number:**
https://jira.eng.vmware.com/browse/VDUCC-46

Signed-off-by: ps37 <prudhvi.af121@gmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "ng": "ng",
     "start": "ng build components; ng build doc-lib; ng serve",
     "build:examples": "ng build examples",
+    "build:examples-aot": "ng build examples --aot=true",
     "build:components": "ng build components",
     "build:doc-lib": "ng build doc-lib",
     "test": "ng test",
@@ -29,13 +30,14 @@
     "@angular/platform-browser-dynamic": "8.2.8",
     "@angular/router": "8.2.8",
     "@clr/angular": "2.2.1",
-    "@clr/ui": "2.2.1",
     "@clr/icons": "2.2.1",
+    "@clr/ui": "2.2.1",
+    "@webcomponents/custom-elements": "1.0.0",
+    "prismjs": "1.17.1",
     "rxjs": "6.4.0",
     "tslib": "1.10.0",
     "zone.js": "0.9.1",
-    "@webcomponents/custom-elements": "1.0.0",
-    "prismjs": "1.17.1"
+    "firebase": ">= 5.5.7 <8"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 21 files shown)


projects/doc-lib/src/doc-lib.module.ts
```diff
@@ -37,11 +38,20 @@ export class DocLibModule {
      * Called in the host package importing this doc library for providing the documentation JSONs needed for
      * {@link CompoDocRetrieverService}
      */
-    public static forRoot(documentations: CompodocSchema[]): ModuleWithProviders {
+    public static forRoot(documentations: any[]): ModuleWithProviders {
```
juanmendes: Why any?

package.json
```diff
@@ -29,13 +30,14 @@
...
     "@clr/icons": "2.2.1",
+    "@clr/ui": "2.2.1",
+    "@webcomponents/custom-elements": "1.0.0",
+    "prismjs": "1.17.1",
     "rxjs": "6.4.0",
     "tslib": "1.10.0",
     "zone.js": "0.9.1",
-    "@webcomponents/custom-elements": "1.0.0",
-    "prismjs": "1.17.1"
+    "firebase": ">= 5.5.7 <8"
```
juanmendes: Firebase is a dev-dependency. Please re-install using save-exact

projects/examples/src/app/app.component.ts
```diff
@@ -25,4 +26,11 @@ export class AppComponent {
         title: entry.displayName,
         path: entry.urlSegment,
     }));
+
+    constructor(router: Router) {
```
juanmendes: Please add an explanation as to why we're doing it here and create an issue to address it later

ps37: Issue: https://jira.eng.vmware.com/browse/VDUCC-72

PR #25, opened 2020-02-05
Update the CI-CD with Pages deployment and npm registry publishing

ps37:
**Why?**
Deploy examples app to Github pages and publish the 2 libraries

**What?**
_Pages deployment:_
Build the examples app in aot mode and use that artifact from the ./dist to deploy. This is being done instead of production build config for building the app because, the prod config is resulting in a package([isuue](https://github.com/angular/angular-cli/issues/11915)) that is not working with the Compodoc JSONs.

_NPM publish:_
The publish stage jobs are conditional based on following conditions:
`($TRAVIS_REPO_SLUG = "vmware/vmware-cloud-director-ui-components") && 
("$TRAVIS_TAG" =~ ^LIB_NAME-v\d.\d.\d$) && 
("$TRAVIS_TAG" != "")`
They should be published to @vcd scope in the registry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,15 @@ cache:
     directories:
         - dist/components
         - dist/doc-lib
+        - dist/examples
 stages:
     - prepare
-    - build-libs
-    - build-app
+    - build-components-lib
+    - build-doc-lib
+    - build-app-aot
     - test
-    - name: publish
-      if: branch = master
+    - publish
+    - deploy
 jobs:
     include:
         - stage: prepare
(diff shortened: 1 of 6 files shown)


juanmendes:
Please add publishing of the doc library and modify the README to explain what we must do when publishing a new version, whether it's a hot patch, or just on top of the latest release

.travis.yml
```diff
@@ -24,20 +26,72 @@ jobs:
...
+          script: skip
+          before_deploy:
+              - cd ./dist/components
+          deploy:
+              edge: true
+              skip_cleanup: true
+              provider: npm
+              email: prudhvi.af121@gmail.com
+              api_key:
+                  secure: P9+lIQscK5k3PAf+r9mhByJIOrbO8/Icx4i2shj+ivkLomXngda4SVn4zI/obkrmNMxPVTpZX+PCq7X8pR1oWETOAiH0TZyVShVxtUGLcwsUSHaE38OPkUZSY+Bcczeavn0s5DrPVKUhBu3ilFv6/5jh6LXLCr0jP8sjdlKo75RN03taefUgM1gCK6amh1b2tgZ2n+1Zo0GQhyX5+yeMqq99U2CmtXtjymjgCM4/DUdH4QL17O4p6eIZqr+rjLAxPif+6C+NLWwRgq2sRHBHjNg0ZRWaAkEO+d938xlDR/upzfV8b9i+vHV8kZwbmwUSNuIIzQfLlFnRMfbLvVONzO5RKRnXhhUU4Ivgdn/BZeir6WGh0hVCxsgSoVq+uaLTjfLhAgtUBr+FL4NEAnFmAdcySgav7eME70Ahd85UF8AvMVFuvbM1F+P5N3Uw7ykGgrcRWWoC6F38pY4wnIR+e81HUYuKCAlb8mUZHJhcfDvfk1lEBWr8ohhDS9SI9rVonl3Enon3NYs/+e+E1MO7v4489XnQ/tLVR4W7ZNbEK/ghhC4AyMNcAVtrgU9+DvJzncvZ/YhlpWiBArAs0jNrmUKug8kh7sEBE1XuFJJUU/lUEfUi3eFeyDnoFbdIyiKfH/euv3zC1l8fbuykfZYzdEzpl0bIw+l4Qtrsbol1YrY=
```
juanmendes: even if encrypted, couldn't anyone just copy/paste this key to their source and run commands with our credentials?

.travis.yml
```diff
@@ -24,20 +26,72 @@ jobs:
...
+        - stage: publish
+          name: Publishing components to npm registry
+          script: skip
+          before_deploy:
+              - cd ./dist/components
+          deploy:
+              edge: true
+              skip_cleanup: true
+              provider: npm
+              email: prudhvi.af121@gmail.com
```
juanmendes: Your personal email? That doesn't sound right...

.travis.yml
```diff
@@ -24,20 +26,72 @@ jobs:
...
           script: npm run build:doc-lib
 
-        - stage: build-app
+        - stage: build-app-aot
           name: Build /examples
-          script: npm run build:examples
+          before_script:
+              - cd ./projects/doc-lib
+              - npm install
+              - npm link
```
juanmendes: We shouldn't need to npm link. I've mentioned it's black magic, and you're not running the two steps that its documentation says are required to get a link working

.travis.yml
```diff
@@ -24,20 +26,72 @@ jobs:
...
+              - cd ../../
+              - npm run build-components-doc
+              - npm run build-examples-doc
+          script: npm run build:examples-aot
+
+        #   TODO: This job is commented out because of failing unit tests of components project
```
juanmendes: Can you please fix the tests that aren't working? If it's too complicated file separate issues for broken tests. Let's not check this in disabling tests

.travis.yml
juanmendes: # Encryption scheme #
Travis CI uses asymmetric cryptography. For each registered repository, Travis CI generates an RSA keypair. Travis CI keeps the private key private, but makes the repository's public key available to those who have access to the repository.

Unless people have access to the Travis CI repository, even though they have the encrypted public key from the file, they cannot get the corresponding private key to publish to npm

PR #29, opened 2020-02-10
Added translation abilities to core components libraries

ryan-bradford:
# Description

These changes provide an i18n framework to localize any application that uses the core components library. We provide the ability for clients to load translations at runtime, or to load them dynamically via an HTTP client. We also took care to make sure that an application using this i18n library can have plugins loaded within it. These plugins will be able to load translations totally independent from the overall application. 

We also provide the default translations for the strings in our core components library. We do not load these strings - that is up to the client. We provide an example script that shows how to merge our provided defaults with your desired translations. 

# Testing

Unit tests were included for all relevant methods/files. We added HTTP translation loading to the examples application and showed that it will work. 

# Usage

- `npm install @vcd/i18n`
- [Import the I18nModule into your app module. If you are a plugin in another application, please use forChild.](https://github.com/ryan-bradford/vmware-cloud-director-ui-components/blob/7ea93e6cc77cb8150e20a9ad63e853fd1ac91751/projects/examples/src/app/app.module.ts#L56)
- Combine your translations with our provided defaults in i18n.json. An example script for doing this can be seen [here](https://github.com/ryan-bradford/vmware-cloud-director-ui-components/blob/7ea93e6cc77cb8150e20a9ad63e853fd1ac91751/projects/i18n/scripts/index.js#L1-L24).
- Either serve these translations as an asset into your application, or import them into your app.module file.
- If you are serving them as an asset, register the asset url as shown [here](https://github.com/ryan-bradford/vmware-cloud-director-ui-components/blob/7ea93e6cc77cb8150e20a9ad63e853fd1ac91751/projects/examples/src/app/app.module.ts#L71-L76) and [provide the asset information to the loaded module ](https://github.com/ryan-bradford/vmware-cloud-director-ui-components/blob/7ea93e6cc77cb8150e20a9ad63e853fd1ac91751/projects/examples/src/app/app.module.ts#L56)
- [Call registerTranslations in the constructor of your app module.](https://github.com/ryan-bradford/vmware-cloud-director-ui-components/blob/7ea93e6cc77cb8150e20a9ad63e853fd1ac91751/projects/examples/src/app/app.module.ts#L81) If you are importing the translations, provided the imported module to this method. 
- Done! Now you can just the translation service as an injected service or as a [pipe](https://github.com/ryan-bradford/vmware-cloud-director-ui-components/blob/7ea93e6cc77cb8150e20a9ad63e853fd1ac91751/projects/examples/src/app/app.component.html#L6).


projects/i18n/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@vcd/ui-i18n",
+  "version": "0.0.1",
+  "peerDependencies": {
+    "@angular/common": "^8.2.9",
```
jmoroski: I think the peer dependency needs to be a range that includes at least the version of Angular 6 that VCD 9.7 supports which is 6.x. 4.4 and beyond would give us compatibility with 9.5.

projects/i18n/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@vcd/ui-i18n",
```
jmoroski: can it just be @vcd/i18n?

projects/i18n/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "version": "0.0.1",
+  "peerDependencies": {
+    "@angular/common": "^8.2.9",
+    "@angular/core": "^8.2.9"
+  },
+  "bin": {
+    "create-translations": "../scripts/index.js"
```
jmoroski: what is the intended usage of this?

juanmendes:
Can you list the steps that users of the library will follow to use this? Like I did for the commit for the doc-lib? Feel free to point to pieces of code in the review with a permalink https://help.github.com/en/github/managing-your-work-on-github/creating-a-permanent-link-to-a-code-snippet

projects/i18n/package.json
ryan-bradford: It was originally going to be a provided script to help plugins combine their translations with our defaults. Now, it is just going to function as an example that I'll talk about in response to Juan's comment. Removed this bin line in the most recent commit.

ryan-bradford:
Updated the pull request with information on how to use the i18n library.

projects/i18n/scripts/index.js
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env node
+const fs = require('fs');
+
+const defaultTranslations = JSON.parse(fs.readFileSync(process.argv[2]).toString());
```
juanmendes: Please add argument documentation

projects/i18n/src/lib/i18n.module.ts
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Creates a {@link I18nModule} using the global translation service.
+     */
+    static forRoot(): ModuleWithProviders {
+        return {
+            ngModule: I18nModule,
+            providers: [
+                {
+                    provide: TranslationService,
+                    useFactory: genericSingletonFactory,
```
juanmendes: I think if you want a single instance, you create it and pass it, why a factory?

projects/i18n/src/lib/i18n.module.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * Creates a {@link I18nModule} using a service that is independent from all other services.
+     * @param extensionRoute the route translations are located at.
+     * @param combined if the translations are in one file or many different files.
+     */
+    static forChild(extensionRoute?: InjectionToken<string>, combined?: boolean): ModuleWithProviders {
+        /**
+         * A implementation of {@link TranslationService} that can inject all of it's dependencies.
```
juanmendes: A => An
it's => its

projects/i18n/src/lib/loader/translation-loader.ts
```diff
@@ -0,0 +1,22 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { HttpClient } from '@angular/common/http';
+import { Observable } from 'rxjs';
+
+export class TranslationLoader {
```
juanmendes: Every exported object should have some docs, even if a single sentence. That also applies to its public members/methods

projects/i18n/src/lib/i18n.module.ts
ryan-bradford: Because we need access to the client's provided `BOOTSTRAP_DETAILS`, which means that the singleton cannot be constructed until the `BOOTSTRAP_DETAILS` are provided.

PR #31, opened 2020-02-11
Added sorting capabilities to datagrid columns

ryan-bradford:
# Description

Exposes information in the gridRefresh event that allows us to sort the columns of the datagrid on the server side. Only supports sorting by 1 column at a time. Supports both ascending and descending sorting. 

# Testing

Added a unit test that tests that a column can be both sorted ascending and descending. Also, added an example that shows a table with two columns and what happens when they are sorted.

sort-columns

# Known Errors

Due to a bug recently fixed in Clarity, when a column is sorted then another is sorted, an icon appears on both columns even though the table is only sorted by the later column.

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -134,22 +134,19 @@ describe('DatagridComponent', () => {
         });
 
         it('displays loading indicators while data is loading', function(this: HasFinderAndGrid): void {
-            spyOn(this.finder.hostComponent, 'refresh').and.callFake(() => {
-                this.finder.detectChanges();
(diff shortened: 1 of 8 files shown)


projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -218,6 +215,47 @@ describe('DatagridComponent', () => {
...
+                    },
+                    {
+                        displayName: 'other',
+                        renderer: 'name',
+                    },
+                ];
+                this.finder.detectChanges();
+            });
+
+            it('should emit the proper column when its sorted', function(this: HasFinderAndGrid): void {
```
juanmendes: Please remove shoulds and proper


Please rework this one so it's under

```ts
describe('@Output() refresh')
```

PR #43, opened 2020-03-02
Implemented grid height + magic pagination

ryan-bradford:
# Description:

Allows the user to set the desired height of their datagrid. If the height is set, the datagrid (all parts included) will be limited to the given height. If the height is not set, the datagrid will fill to expand the parent. If the parents height is not fixed, the datagrid will grow to the necessary size.

Also, added the "Magic" page size option. This will calculate the optimal page size given the height of the datagrid. If the height is not set, it will not attempt to calculate the page size, and will instead default to 5.

# Testing:

Added unit tests to test both the grid height input and the "Magic" pagination setting. Also, added an example to show the datagrid filling up the whole parent container, and an example showing the datagrid being limited to it's set height and calculating the correct page size amount.

Screen Shot 2020-03-02 at 3 15 30 PM

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -227,15 +227,89 @@ describe('DatagridComponent', () => {
                 });
             });
 
+            describe('@Input() pagination', () => {
+                describe('pageSize', () => {
+                    it('magic pagination finds the most rows that can fit in the set height', function(this: HasFinderAndGrid): void {
+                        this.finder.hostComponent.height = 200;
+                        this.finder.detectChanges();
+                        this.finder.hostComponent.pagination = {
+                            pageSize: 'Magic',
+                            pageSizeOptions: [10],
+                        };
+                        this.finder.detectChanges();
+                        expect(this.clrGridWidget.getPaginationDescription()).toEqual(' 1 - 3 of 150 items ');
+                    });
+
+                    it('magic pagination does nothing when height is not set', function(this: HasFinderAndGrid): void {
+                        this.finder.hostComponent.pagination = {
+                            pageSize: 'Magic',
+                            pageSizeOptions: [10],
+                        };
+                        this.finder.detectChanges();
+                        expect(this.clrGridWidget.getPaginationDescription()).toEqual(' 1 - 5 of 150 items ');
+                    });
+
+                    it('lets the user set rows per page', function(this: HasFinderAndGrid): void {
+                        this.finder.hostComponent.pagination = {
+                            pageSize: 100,
+                            pageSizeOptions: [10],
+                        };
+                        this.finder.detectChanges();
+                        expect(this.clrGridWidget.getPaginationDescription()).toEqual(' 1 - 100 of 150 items ');
+                    });
+
+                    it('creates a smaller page when buttons are present', function(this: HasFinderAndGrid): void {
+                        this.finder.hostComponent.height = 200;
+                        this.finder.detectChanges();
+                        this.finder.hostComponent.pagination = {
+                            pageSize: 'Magic',
+                            pageSizeOptions: [10],
+                        };
+                        this.finder.hostComponent.buttonConfig = {
+                            globalButtons: [
+                                {
+                                    label: 'Add',
+                                    isActive: () => true,
+                                    handler: () => {},
+                                    class: 'button',
+                                },
+                            ],
+                            contextualButtonConfig: {
+                                buttons: [],
+                                featured: [],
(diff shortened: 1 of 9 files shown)


projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -660,6 +736,8 @@ export class HostWithDatagridComponent {
 
     selectionType = GridSelectionType.None;
 
+    height: number | undefined = undefined;
```
juanmendes: Can't this just be `height: number;`?

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -674,6 +752,14 @@ export class HostWithDatagridComponent {
 
     paginationText = 'Total Items';
 
+    pagination: {
```
juanmendes: Please create a definition for it, explaining what each thing means. Let's make it all lower case so it doesn't look like a string that could be displayed. Didn't we say we would have a way for the user to specify the height of each row?

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -28,6 +30,26 @@ import {
 import { ContextualButton } from './interfaces/datagrid-column.interface';
 import { ComponentRendererSpec } from './interfaces/component-renderer.interface';
 
+/**
+ * The default number of items on a single page.
+ */
```
juanmendes: Let's change this default to 10, 5 is too few

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -535,6 +585,61 @@ export class DatagridComponent<R> implements OnInit {
         return this.paginationCallback(paginationData.firstItem + 1, paginationData.lastItem + 1, this.totalItems);
     }
 
+    /**
+     * The number of rows in a single page.
+     */
+    private getPageSize(): number {
+        if (typeof this.pagination.pageSize === 'number') {
```
juanmendes: Please remove elses

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -511,21 +565,17 @@ export class DatagridComponent<R> implements OnInit {
     }
 
     /**
-     * Resets the pagination to page 1.
+     * Is the given column hideable?
```
juanmendes: Probably better to explain what hideable means instead of repeating the text of the method

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -218,6 +242,8 @@ export class DatagridComponent<R> implements OnInit {
         return this._buttonConfig;
     }
 
+    constructor(private component: ElementRef) {}
```
juanmendes: component is misleading, it's an html element, a node. A component is an abstraction over a node

projects/components/src/datagrid/datagrid.component.spec.ts
ryan-bradford: I think I did what you were asking for here, let me know if it isn't. Also, totally forgot about the row height - added that in the latest commit (and a test for it).

ryan-bradford:
Updated per review!

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -240,13 +240,25 @@ describe('DatagridComponent', () => {
...
+                    it('allows the user to set a custom row height with magic pagination ', function(this: HasFinderAndGrid): void {
+                        this.finder.hostComponent.height = 200;
+                        this.finder.detectChanges();
+                        this.finder.hostComponent.pagination = {
+                            pageSize: 'Magic',
+                            pageSizeOptions: [10],
+                            rowHeight: 100,
+                        };
+                        this.finder.detectChanges();
+                        expect(this.clrGridWidget.getPaginationDescription()).toEqual(' 1 - 1 of 150 items ');
```
juanmendes: Can we fix the HTML so it doesn't have spaces?

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -273,45 +296,16 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, AfterViewChe
...
-    } {
+    get pagination(): PaginationConfiguration {
         return this._pagination;
     }
 
-    _pagination: {
-        pageSizeOptions: number[];
-        pageSize: number | 'Magic';
-    } = {
+    _pagination: PaginationConfiguration = {
```
juanmendes: Should be private

PR #47, opened 2020-03-09
Adds activity reporters Into core components

ryan-bradford:
# Description

Adds the activity reporter from VCD-UI into the core components library. This abstract class allows for components to monitor observables and post errors when appropriate. This change also required moving over the Error Banner and the Loading Indicator. 

I removed the dependency to ResovledCursor by extracting an interface (`ActivityResponse`) that is the simplest set of information we need (just error? and status). After that, everything was mostly copy-paste.

# Testing

Added unit tests to achieve ~100% coverage of the diff. In addition, added examples of the loading indicator, the error banner, and both types of activity reporters (banner + spinner)

activity-monitor

# Notes to Reviewer

Some items that I would like a second opinion on:
1. In the tests for banner and spinner activity reporter, I had to disable the line length checker for the it statements. I couldn't find another way to rework it so that the line was not too long, but open to suggestions.
2. In the activity-reporter.examples.module.ts file, I had to use `BannerActivityReporterComponent` as the component for the documentation registration. This is because `ActivityReporter` is an abstract class that doesn't satisfy the requirements for `component`. I put a TODO requesting a change to doc-lib to support documentation for non-component files, but I don't know if there is a better way.

### DIFF
--- a/projects/components/src/common/activity-reporter/activity-reporter.spec.ts
+++ b/projects/components/src/common/activity-reporter/activity-reporter.spec.ts
@@ -1,0 +1,70 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { ActivityReporter } from './activity-reporter';
+
+interface HasPromiseAndReporter {
+    // The data about the promise for {@link ActivityReporter.monitorEdit}.
+    promise: Promise<string>;
+    // The function that will reject the promise.
+    reject: (error: string) => void;
+    // The function that will resolve the promise.
+    resolve: (resolved: string) => void;
+    // Spys on {@link ActivityReporter.start}
+    startSpy: jasmine.Spy;
+    // Spys on {@link ActivityReporter.reporterError}
+    reportErrorSpy: jasmine.Spy;
+    // Spys on {@link ActivityReporter.reporterSuccess}
+    reportSuccessSpy: jasmine.Spy;
+    // The activity reporter used for testing.
+    reporter: ActivityReporter;
+}
+
+describe('ActivityReporter', () => {
+    beforeEach(function(this: HasPromiseAndReporter): void {
+        this.promise = new Promise<string>((promiseResolve, promiseReject) => {
+            this.resolve = stuff => promiseResolve(stuff);
+            this.reject = stuff => promiseReject(stuff);
+        });
+
+        this.reporter = new SimpleReporter();
+        this.startSpy = spyOn(this.reporter, 'startActivity');
+        this.reportErrorSpy = spyOn(this.reporter, 'reportError');
+        this.reportSuccessSpy = spyOn(this.reporter, 'reportSuccess');
+    });
+
+    class SimpleReporter extends ActivityReporter {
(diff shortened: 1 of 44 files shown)


projects/components/src/common/activity-reporter/activity-reporter.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    class SimpleReporter extends ActivityReporter {
+        reportError(errorText: string): void {}
+
+        reportSuccess(successMessage?: string): void {}
+
+        startActivity(): void {}
+    }
+
+    describe('monitorActivity', () => {
+        it('can monitor an activity with success', done => {
```
juanmendes: Please improve the text, when we see the same text in multiple its, it's a sign. "can monitor an activity"

I would expect test cases that say

it displays a default message while the promise is pending
it displays a custom message while the promise is pending
it removes the activity reporter from the screen after the promise resolves
it displays an error on the screen if the promise is rejected

And maybe the same thing for when there are multiple promises (`monitorActivities `).

We may need to chat about this when I come back

projects/components/src/common/activity-reporter/activity-reporter.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+                expect(reportSuccess).toHaveBeenCalledTimes(1);
+                done();
+            });
+
+            resolve({
+                status: ActivityStatus.READY,
+            });
+        });
+
+        it('can monitor an activity resolved failure', done => {
```
juanmendes: Couldn't we use async await? Or at least return a Promise? done was created before we had Promises and async/await.

projects/components/src/common/activity-reporter/banner-activity-reporter.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import { BannerActivityReporterWidgetObject } from '../../utils/test/activity-reporter/banner-activity-reporter.wo';
+import { TranslationService, MockTranslationService } from '@vcd/i18n';
+
+@Component({
+    template: `
+        <vcd-banner-activity-reporter #activityReporter></vcd-banner-activity-reporter>
+    `,
+})
+class TestBannerComponent {
+    @ViewChild('activityReporter', { static: true }) activityReporter: BannerActivityReporterComponent;
```
juanmendes: You don't need a hash if there's only one `vcd-banner-activity-reporter` See https://angular.io/api/core/ViewChild#example-2

projects/components/src/common/activity-reporter/activity-reporter.spec.ts
ryan-bradford: Reworked the messages a little in both this spec file and the spec file for the two concrete ActivityReporters. Let me know what you think

ryan-bradford:
Updated per today's conversation in the standup. Only thing is I ended up removing `monitorActivities` because it was only used in one place in VCD-UI. The purpose of the method wasn't to take in a list of promises, but a single promise that returns a list of statuses. It then checked the statuses to determine if all of them didn't error. Because we aren't supporting status, it didn't make sense to keep this method.

PR #48, opened 2020-03-10
Add components for data grid filters with selection inputs

ps37:
### Following are a list of filters with selection input:
Single select filter
Multi-select filter
Single select filter with custom FIQL

### Single select filter example:
selectFilter

### Multi-select filter example:
multiSelectFilter

### DIFF
--- a/projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
+++ b/projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
@@ -1,0 +1,110 @@
+/*!
+ * Copyright 2019 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { Component, Host, Input, OnInit } from '@angular/core';
+import { DatagridFilter, FilterConfig } from './datagrid-filter';
+import { SelectOption } from './datagrid-select-filter.component';
+import { ClrDatagridFilter } from '@clr/angular';
+import { FormControl, FormGroup } from '@angular/forms';
+import { FilterBuilder } from '../../utils/filter-builder';
+
+/**
+ * Same as the {@link SelectOption} but value is always a string
+ */
+export interface MultiSelectOption extends SelectOption {
+    value: string;
+}
+
+export interface DatagridMultiSelectFilterConfig extends FilterConfig<string[]> {
+    /**
+     * List of select options
+     */
+    options: MultiSelectOption[];
+}
+
+/**
+ * Used within a clarity grid column {@link DatagridComponent} to render a filter widget with list of checkboxes to select
+ * multiple options
+ * Example usage:
+ * columns: GridColumn<MockRecord>[] = [{
+ *      displayName: 'Custom Select filter',
+ *      renderer: 'state',
+ *      queryFieldName: 'state',
+ *      filterRendererSpec: FilterComponentRendererSpec({
+ *          type: DatagridMultiSelectFilterComponent,
+ *          config: {
+ *              options: [{
+ *                  value: 'CA',
+ *                  display: 'California'
+ *              }, {
+ *                  value: 'MA',
+ *                  display: 'Massachusetts'
+ *              }],
+ *              value: ['MA', 'NC'],
+ *          },
+ *      }),
+ *   }]
+ */
+@Component({
+    selector: 'vcd-dg-multiselect-filter',
+    templateUrl: 'datagrid-multiselect-filter.component.html',
+})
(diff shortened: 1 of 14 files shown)


projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
```diff
@@ -0,0 +1,107 @@
...
+        return new FilterBuilder().any(...selectedFilters).getString();
+    }
+
+    isActive(): boolean {
+        return (
+            this.formGroup &&
+            !!Object.keys(this.formGroup.getRawValue()).filter(frmCtrl => this.formGroup.get(frmCtrl).value).length
+        );
+    }
+}
```
ryan-bradford: Can you add a factory function to these filters like in `DatagridNumericFilter`? Also, how will this interact with the dynamic options?

projects/components/src/datagrid/filters/datagrid-select-filter.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    constructor(@Host() private filterContainer: ClrDatagridFilter, private fb: FormBuilder) {
+        super(filterContainer);
+        this.formGroup = this.fb.group({
+            filterSelect: '',
+        });
+    }
+
+    ngOnInit(): void {
```
juanmendes: Do we really need to debounce a select element?

projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
```diff
@@ -0,0 +1,107 @@
...
+        if (this._config.value) {
+            this.setValue(this._config.value);
+        }
+    }
+    get config(): DatagridMultiSelectFilterConfig {
+        return this._config;
+    }
+
+    ngOnInit(): void {
+        this.debounceChanges(this.formGroup.valueChanges);
```
juanmendes: Do we really need to debounce checkboxes? Typing makes sense, but most people will take at least 500ms between clicks

projects/components/src/utils/test/datagrid/filter-utils.ts
```diff
@@ -15,6 +15,7 @@ import {
     GridDataFetchResult,
 } from '../../../datagrid';
 import { MockRecord } from '../../../datagrid/mock-data';
+import { MockTranslationService, TranslationService } from '@vcd/i18n';
```
juanmendes: Please have imports from external packages be at the top

projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    getValue(): string {
+        const selectedFilters = Object.keys(this.formGroup.getRawValue())
+            .filter(frmCtrl => this.formGroup.get(frmCtrl).value)
+            .map(selectedOption => new FilterBuilder().is(this.queryField).equalTo(selectedOption));
+        return new FilterBuilder().any(...selectedFilters).getString();
+    }
+
+    isActive(): boolean {
+        return (
```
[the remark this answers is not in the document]
ps37: That's Prettier adding the paranthesis

projects/components/src/utils/id-generator/id-generator.ts
```diff
@@ -0,0 +1,26 @@
+/*!
+ * Copyright 2019 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/**
+ * Object containing the logic to generate unique identifiers whenever it's {@link IdGenerator#id} method is called
```
juanmendes: Object containing is redundant. Please avoid words like class, object, interface, logic, code, called in documentation.

Generates unique IDs.

projects/components/src/utils/id-generator/id-generator.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export class IdGenerator {
+    /**
+     * Variable that is incremented whenever a new ID is needed
```
juanmendes: No need for this comment. private variables should be documented if they are confusing.

projects/components/src/utils/id-generator/id-generator.ts
```diff
@@ -0,0 +1,26 @@
...
+     */
+    private static idCounter = 0;
+
+    /**
+     * Getter that returns the a unique ID
```
juanmendes: No need to say getter, the code already says that. Having said that, a getter that always returns a different value seems wrong. Please rename to `generate` following the pattern from https://mongodb.github.io/mongo-java-driver/3.8/javadoc/org/bson/codecs/IdGenerator.html

PR #51, opened 2020-03-12
Factories for column renderers and filters.

ryan-bradford:
# Description

Adds factory methods to the filter components and the column renderers. Interface is very simple and just a single static method.

# Testing

Used the factories in the examples app in-place of the current configuration. Also, added unit tests for the filter factories. I **did not add** unit tests for the column render factories because there is not a spec file for either the `BoldTextRendererComponent` or `LinkedTextRendererComponent`. I will add these if it's suggested. 


projects/components/src/datagrid/filters/datagrid-numeric-filter.component.ts
```diff
@@ -26,15 +26,26 @@ export type DatagridNumericFilterConfig = FilterConfig<[number, number]>;
...
             [FormFields.from]: null,
             [FormFields.to]: null,
         });
     }
+    formGroup: FormGroup;
+
+    /**
+     * Creates a {@link FilterRendererSpec} with the given config.
+     */
+    static factory(value?: [number, number]): FilterRendererSpec<DatagridNumericFilterConfig> {
```
juanmendes: Can we rename this to create?

Also let's create a named type for this tuple. We should do that if we find ourselves repeating an inline type.

```typescript
type Range = [number, number]
```

projects/components/src/datagrid/filters/datagrid-string-filter.component.ts
```diff
@@ -35,6 +35,22 @@ export class DatagridStringFilterComponent extends DatagridFilter<string, Datagr
         });
     }
 
+    /**
+     * Creates a {@link FilterRendererSpec} with the given config.
+     */
+    static factory(
```
juanmendes: Please rename to `create` and specify `@param` in the docs

projects/components/src/datagrid/filters/datagrid-numeric-filter.component.spec.ts
```diff
@@ -2,20 +2,23 @@
...
+import { DatagridFilter, DEBOUNCE_TIME_FOR_GRID_FILTER_CHANGES } from './datagrid-filter';
+import { DatagridNumericFilterConfig, DatagridNumericFilterComponent } from './datagrid-numeric-filter.component';
 
 interface HasDgNumericFilter {
     filter: DatagridFilter<[number, number], DatagridNumericFilterConfig>;
 }
 
 describe('Datagrid numeric filter', () => {
+    describe('create', () => {
+        it('can create the filter via factory', () => {
```
juanmendes: I like to make the text of the it talk about its functionality

```ts
it('simplifies the creation of filters')
```

projects/components/src/datagrid/renderers/linked-text-renderer.component.ts
```diff
@@ -5,6 +5,7 @@
```
juanmendes: This can needs some work until it's actually useful. It's not worth maintaining it (adding unit tests). Let's remove this, and add unit tests for the `BoldTextRenderer` which can be used as an example until we have more useful renderers.

projects/components/src/datagrid/filters/datagrid-string-filter.component.ts
ryan-bradford: Added documentation for params. But the reason I didn't name it `create` is because any class that extends SubscriptionTrackerMixin(Object) inherits all of Objects static methods. Object comes with a `create` method that clashes when I try and add this `create` method. Totally open to other names however.

juanmendes: Let's export a separate function that is named by removing the `Component` suffix

`DatagridStringFilter(params)`

PR #52, opened 2020-03-13
Implemented auto cliptext in the grid cells

ryan-bradford:
# Description

Grid cells will automatically enable cliptext using the show-cliptext directive. This is always the case unless the client sets disableCliptext to true within the column configuration.

# Testing

Added an example that shows text being clipped for normal renderers and component renderers. It also shows cliptext being disabled. Also, added tests to test if disabling cliptext actually removes the directive. 

cliptext

# Reviewer Notes:
1. I chose to make cliptext size be always large. Should this be an input option?

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.html
+++ b/projects/components/src/datagrid/datagrid.component.html
@@ -111,7 +111,7 @@
             </div>
         </clr-dg-cell>
 
-        <clr-dg-cell *ngFor="let column of columnsConfig">
+        <clr-dg-cell [vcdShowClippedText]="column.cliptextConfig" *ngFor="let column of columnsConfig">
             <!-- Default renderer -->
             <ng-container *ngIf="column.fieldName">{{ restItem | nestedProperty: column.fieldName }}</ng-container>
 
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -3,25 +3,26 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { Component, ViewChild, TemplateRef, ContentChild } from '@angular/core';
+import { Component, ContentChild, TemplateRef, ViewChild } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import { TooltipSize } from '../lib/directives/show-clipped-text.directive';
+import { ClrDatagridWidgetObject } from '../utils/test/datagrid/datagrid.wo';
+import { WidgetFinder } from '../utils/test/widget-object';
 import { GridSelectionType, PaginationConfiguration } from './datagrid.component';
 import { DatagridComponent, GridDataFetchResult, GridState } from './datagrid.component';
+import { DatagridModule } from './datagrid.module';
 import {
+    ButtonConfig,
+    ColumnComponentRendererSpec,
+    ContextualButtonPosition,
     GridColumn,
     GridColumnHideable,
-    ButtonConfig,
-    ContextualButtonPosition,
     InactiveButtonDisplayMode,
-    ColumnComponentRendererSpec,
 } from './interfaces/datagrid-column.interface';
-import { TestBed } from '@angular/core/testing';
-import { DatagridModule } from './datagrid.module';
-import { WidgetFinder } from '../utils/test/widget-object';
-import { BoldTextRendererComponent } from './renderers/bold-text-renderer.component';
-import { ClrDatagridWidgetObject } from '../utils/test/datagrid/datagrid.wo';
-import { WithGridBoldRenderer } from './renderers/bold-text-renderer.wo';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { mockData, MockRecord } from './mock-data';
+import { BoldTextRendererComponent } from './renderers/bold-text-renderer.component';
+import { WithGridBoldRenderer } from './renderers/bold-text-renderer.wo';
 
 type MockRecordDatagridComponent = DatagridComponent<MockRecord>;
 
@@ -108,6 +109,38 @@ describe('DatagridComponent', () => {
                     'secondRowB',
                     'Expected a different new class to display for the second row.'
                 );
+            });
+
+            describe('@Input() columns.disableCliptext', () => {
+                beforeEach(function(this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.gridData = {
+                        items: mockData,
+                        totalItems: 2,
+                    };
+                    this.finder.detectChanges();
+                });
+
+                it('clips text when disableCliptext is unset', function(this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.columns = [{ displayName: 'Name', renderer: 'name' }];
+                    this.finder.detectChanges();
+                    expect(this.clrGridWidget.columnClippedTextDirective(0).disabled).toBeFalsy();
+                });
+
+                it('clips text when disableCliptext is false', function(this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.columns = [
+                        { displayName: 'Name', renderer: 'name', cliptextConfig: { size: TooltipSize.md } },
+                    ];
+                    this.finder.detectChanges();
+                    expect(this.clrGridWidget.columnClippedTextDirective(0).disabled).toBeFalsy();
+                });
+
+                it('does not clip text when disableCliptext is true', function(this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.columns = [
+                        { displayName: 'Name', renderer: 'name', cliptextConfig: { disabled: true } },
+                    ];
+                    this.finder.detectChanges();
+                    expect(this.clrGridWidget.columnClippedTextDirective(0).disabled).toBeTruthy();
+                });
             });
 
             describe('@Input() selectionType', () => {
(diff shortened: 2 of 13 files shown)


juanmendes:
Yes, we should provide the option for the user to select the size and the timeout for the tooltip.

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -35,6 +35,15 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
         return this.getNodeText(this.getCell(row, column));
     }
 
+    /**
```
juanmendes: Since there's no way for two rows on the same column to return a different value, this should not take the row index as an input. It also doesn't really test is going to be clipped, which is a good thing because this class is not responsible for that. Let's rename it to 

```typescript
columnHasClippedTextDirective(column: number): boolean
```

We should try querying for the directive instead of using what seems to be an internal feature, it's an array of `any` without much documentation.

projects/components/src/datagrid/datagrid.component.html
```diff
@@ -111,22 +111,30 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
             </div>
         </clr-dg-cell>
 
-        <clr-dg-cell *ngFor="let column of columnsConfig">
-            <!-- Default renderer -->
-            <ng-container *ngIf="column.fieldName">{{ restItem | nestedProperty: column.fieldName }}</ng-container>
+        <div *ngFor="let column of columnsConfig">
+            <ng-template #cellTemplate>
```
juanmendes: If we can have an enabled switch on the clipped text directive, we won't need to dynamically add the directive, removing the duplication of HTML?

projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
```diff
@@ -234,6 +234,11 @@ export interface GridColumn<R> {
      * and sorting feature on. And we want filtering to be off on some columns while still having sorting enabled.
      */
     sortBy?: string;
+
+    /**
+     * When disableCliptext is true, the cells within the datagrid will not be clipped.
+     */
+    disableCliptext?: boolean;
```
juanmendes: This should take options to be passed the the directive, one of them could be the option to disable it, as mentioned in a different comment.

ryan-bradford:
Updated per the review. Some sketchy stuff to take a look at:

1. In datagrid.component.ts I map over columns in `set columns`, but this might mutate the underlying column object? Not sure what to do here. 
2. How dynamic disabled is implemented seems to work, but needs a second pair of eyes

projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
```diff
@@ -236,9 +251,11 @@ export interface GridColumn<R> {
     sortBy?: string;
 
     /**
-     * When disableCliptext is true, the cells within the datagrid will not be clipped.
+     * The configuration for the cliptext in the datagrid.
+     * Defaults to size: 'lg', mouseoutDelay: undefined.
+     * If null, will disable cliptext
      */
-    disableCliptext?: boolean;
+    cliptextConfig?: CliptextConfig | null;
```
juanmendes: If it's optional, no need to make it `| null`

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -233,6 +239,9 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
     ContextualButtonPosition = ContextualButtonPosition;
     GridColumnHideable = GridColumnHideable;
     TooltipSize = TooltipSize;
+    DEFAULT_CLIPTEXT: CliptextConfig = {
```
juanmendes: Does this really need to be a property on the class? I think just an object outside the class is fine, and it won't require `this.`

projects/components/src/lib/directives/show-clipped-text.directive.spec.ts
```diff
@@ -120,6 +121,18 @@ describe('ShowClippedTextDirective', () => {
         // });
     });
 
+    describe('disabling the tooltip', () => {
+        it('wont show a tooltip when disabled', fakeAsync(function(this: Test): void {
+            const helper = this.clippedTextHelper;
+            helper.disabled = true;
```
juanmendes: I think the disabled setter should call detect changes, then we wouldn't need to expose the fixture and we'd avoid some duplication

projects/components/src/lib/directives/show-clipped-text.directive.ts
```diff
@@ -209,6 +209,23 @@ export class ShowClippedTextDirective implements OnDestroy, OnInit {
     @Input('vcdShowClippedTextMouseOutDelay')
     mouseoutDelay = 500;
 
+    @Input('vcdShowClippedTextDisabled')
+    set disabled(disabled: boolean) {
+        if (disabled && this.init) {
+            this.ngOnDestroy();
```
juanmendes: I think we should have methods that initialize and remove the handlers instead of calling `ngOnDestroy` and `ngOnit` manually. They are only supposed to be called by the framework. Let's chat about it before you try to rework it

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -4,15 +4,13 @@
...
-/*!
- * Copyright 2019 VMware, Inc.
- * SPDX-License-Identifier: BSD-2-Clause
- */
 
 import { WidgetObject } from '../widget-object';
 import { DebugElement, Type } from '@angular/core';
 import { ClrDatagrid } from '@clr/angular';
 import { DatagridFilter } from '../../../datagrid';
+import { By } from '@angular/platform-browser';
```
juanmendes: Please reorder the imports so that imports from `@anything` are at the top

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -37,11 +35,11 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
 
     /**
      * Retrieves if the cell will clip text
-     * @param row 0-based index of row
      * @param column 0-based index of column
      */
-    cellWillClip(row: number, column: number): boolean {
-        return this.getCell(row, column).providerTokens.some(token => token.name === 'ShowClippedTextDirective');
+    columnHasClippedTextDirective(column: number): boolean {
```
juanmendes: Since we are always adding the directive, this name doesn't make sense anymore...

ryan-bradford:
Updated per the conversation this morning + review!

I had an error that the build failed if the datagrid tests ran before the cliptext tests. I think I solved this error (build seems stable again), but just to be aware.

projects/components/src/datagrid/datagrid.component.html
```diff
@@ -111,22 +111,28 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
             </div>
         </clr-dg-cell>
 
-        <clr-dg-cell *ngFor="let column of columnsConfig">
-            <!-- Default renderer -->
-            <ng-container *ngIf="column.fieldName">{{ restItem | nestedProperty: column.fieldName }}</ng-container>
+        <div *ngFor="let column of columnsConfig">
+            <clr-dg-cell
+                [vcdShowClippedText]="getCliptextSize(column.cliptextConfig)"
```
juanmendes: Why isn't this a single config object?

```
[vcdClippedTextConfig]="getClipTextConfig(column)"
```

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -110,6 +111,33 @@ describe('DatagridComponent', () => {
                 );
             });
 
+            describe('@Input() columns.disableCliptext', () => {
+                beforeEach(function(this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.columns = [
```
juanmendes: You can probably configure this in each `it` separately. Then you don't need magic indexes in each test

projects/components/src/datagrid/datagrid.component.html
```diff
@@ -111,22 +111,24 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
             </div>
         </clr-dg-cell>
 
-        <clr-dg-cell *ngFor="let column of columnsConfig">
-            <!-- Default renderer -->
-            <ng-container *ngIf="column.fieldName">{{ restItem | nestedProperty: column.fieldName }}</ng-container>
+        <div *ngFor="let column of columnsConfig">
```
juanmendes: Why the wrapping div? If really needed, it should be an `ng-container` to avoid adding an HTML element but I thought you should be able to use it in the same tag as the *ngFor

```
<clr-dg-cell *ngFor="let column of columnsConfig; let column = column" [vcdShowClippedText]="column.cliptextConfig">

```

projects/components/src/datagrid/datagrid.component.html
```diff
@@ -111,22 +111,24 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
...
-            <!-- Renderer is a componentRenderer -->
-            <ng-template
-                *ngIf="column.fieldColumnRendererSpec"
-                [vcdComponentRendererOutlet]="{ rendererSpec: column.fieldColumnRendererSpec, context: restItem }"
-            >
-            </ng-template>
-        </clr-dg-cell>
+                <!-- Renderer is a componentRenderer -->
+                <ng-template
+                    *ngIf="column.fieldColumnRendererSpec"
```
juanmendes: Please move the `*ngIf` to the same line as `<ng-template>`, I know it's not your doing but that's how we typically do it

PR #54, opened 2020-03-19
Enabled activity reporting on the grid

ryan-bradford:
# Description

Added the activity reporter on top of the grid. Exposes it as a field of the datagrid, and also allows button handlers to return a promise that triggers the activity reporting. 

# Testing

Added unit tests to test that it works with the datagrid. In addition, added an example to show how to return a promise from the button handler and trigger reporting.

# Reviewer Notes:
1. Datagrid example is a little abstract.
2. Searching this.root.parent for h3 in the datagrid.wo.ts file to find the header feels strange.
3. Should we provide a better export from the datagrid to monitor activities outside of the grid buttons?

grid-activity

### DIFF
--- a/projects/components/src/utils/test/datagrid/datagrid.wo.ts
+++ b/projects/components/src/utils/test/datagrid/datagrid.wo.ts
@@ -200,8 +200,8 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
      * Gives the header above the grid.
      */
     get gridHeader(): string | undefined {
-        const headerElement = this.root.nativeElement.previousSibling;
-        return headerElement.nodeType !== 8 ? headerElement.textContent : undefined;
+        const headerElement = this.findElement('h3', this.root.parent);
+        return headerElement ? headerElement.nativeElement.textContent : undefined;
     }
 
     /**
--- a/projects/components/src/utils/test/datagrid/datagrid.wo.ts
+++ b/projects/components/src/utils/test/datagrid/datagrid.wo.ts
@@ -6,7 +6,6 @@
 import { GridSelectionType } from './../../../datagrid/datagrid.component';
 
 import { DebugElement, Type } from '@angular/core';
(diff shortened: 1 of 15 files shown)


projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -200,8 +200,8 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
      * Gives the header above the grid.
      */
     get gridHeader(): string | undefined {
-        const headerElement = this.root.nativeElement.previousSibling;
-        return headerElement.nodeType !== 8 ? headerElement.textContent : undefined;
+        const headerElement = this.findElement('h3', this.root.parent);
```
juanmendes: As you mentioned, this feels weird because it is wrong. This widget object is only supposed to wrap the ClrDatagrid. Our Grid contains extra stuff, we really need a separate DataGrid object since it contains a `ClrDataGrid`

Let me know if you'd like to chat about this. I think it just needs to have a public method to get its clr data grid and one more method to get the grid header.

This looks like it should be defined using `getText`

ryan-bradford:
Split the ClrDatagrid and the VcdDatagrid into two (related) widget objects!

projects/components/src/utils/test/datagrid/vcd-datagrid.wo.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Widget Object for our VCD DataGrid
+ */
+export class VcdDatagridWidgetObject<R> extends WidgetObject<DatagridComponent<R>> {
+    static tagName = `vcd-datagrid`;
+
+    /**
+     * Gives the header above the grid.
+     */
+    get gridHeader(): string | undefined {
+        return this.findElement('h3') ? this.getText('h3') : undefined;
```
juanmendes: Can we instead improve `getText` so that it returns an empty string if a null element is passed to it?

PR #58, opened 2020-03-23
Fix the master branch condition expression for CI jobs

ps37:
The if condition expressions to perform jobs only on the master branch were wrong and the condition expressions are corrected.
It didn't fail during my testing because of the `on:` condition for the workflow at top of the yaml file. Previously, it was only running on pushes to master branch and I changed it just before finalizing my changes and didn't test the condition.

Test Run: https://github.com/ps37/vmware-cloud-director-ui-components/runs/528355791?check_suite_focus=true


ps37:
> Can you also bump components version in package json?

I bumped it

PR #60, opened 2020-03-26
Add not sortable flag to grid column

ps37:
As part of this PR, also the following changes are added:

Remove default filter

Make the subscription tracker mixin extent 'class {}' instead of Object

**Demo:**
notSortableFlag

### DIFF
--- a/projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
+++ b/projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
@@ -224,13 +224,7 @@ export interface GridColumn<R> {
      * TODO: Should this be made to work with top level search on grids across all columns?
      *  The above to-do is going to be worked on as part of https://jira.eng.vmware.com/browse/VDUCC-27 and
      */
-    filterRendererSpec?: FilterRendererSpec<FilterConfig<unknown>>;
-
-    /**
-     * To enable only sorting without filtering on the column. This is because, passing in clrDgField turns both filtering
-     * and sorting feature on. And we want filtering to be off on some columns while still having sorting enabled.
-     */
-    sortBy?: string;
+    filter?: FilterRendererSpec<FilterConfig<unknown>>;
 
     /**
      * The configuration for the cliptext in the datagrid.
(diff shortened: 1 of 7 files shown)


juanmendes:
Please add tests/examples for sortable. We should also have tests/examples for the filters not displaying.

projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
```diff
@@ -224,20 +224,20 @@ export interface GridColumn<R> {
...
      * Defaults to size: 'lg', mouseoutDelay: undefined.
      * If null, will disable cliptext
      */
     cliptextConfig?: CliptextConfig;
+
+    /**
+     * Boolean to turn off sorting on the column. Adding a {@link GridColumn.queryFieldName} by default turns both
+     * filtering and sorting on and this switch allows for turning on just the filtering
+     */
+    notSortable?: boolean;
```
juanmendes: Can this be called sortable instead? `!notSortable` looks weird. To default it consider adding a getter that tests if the property exists

```ts
/**
 *  Whether a columns is sortable, defaults to true
 */
isColSortable(col: SomeType) {
   return 'sortable' in col ? col.sortable : true; 
  // or
  return !(col.sortable === false)
}
```

projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
```diff
@@ -224,20 +224,20 @@ export interface GridColumn<R> {
...
 
     /**
      * The configuration for the cliptext in the datagrid.
      * Defaults to size: 'lg', mouseoutDelay: undefined.
      * If null, will disable cliptext
      */
     cliptextConfig?: CliptextConfig;
+
+    /**
+     * Boolean to turn off sorting on the column. Adding a {@link GridColumn.queryFieldName} by default turns both
```
juanmendes: Again, don't use words that the code already specifies.

Adding a query field name does not turn filtering on

`Whether to show the sorting toggle. Defaults to true`

ps37:
> Do we need to update any unit tests?

Added them to datagrid component spec

PR #63, opened 2020-03-27
Fixed cliptext tests.

ryan-bradford:
# Description

Fixes the cliptext tests so that they always pass, regardless of if the window is focused or it is running in GH Actions.

Necessary Changes made:

1. transitionend doesn't seem to run when the window is unfocused. So I made the tip static object exported (but not exported from the public API) and directly called the transitionEnd method from the test.
2. mutationObserver threw an error that hostElement could not be found because tip.currentDirective is undefined. So, I added a check to make sure tip.currentDirective is defined.

Those were the two issues causing the test failures.

Improvements Made:

1. Removed most of the await timeout(...) to speed up the tests. Only two were required to still be there and are both commented (and we're commented originally before this change). 

# Testing

Ran the tests locally numerous times. Also, ran them on GH Actions a dozen or so times by creating pull requests and pushing empty commits. They appear to always pass now.

### DIFF
--- a/projects/components/src/lib/directives/show-clipped-text.directive.spec.ts
+++ b/projects/components/src/lib/directives/show-clipped-text.directive.spec.ts
@@ -4,7 +4,7 @@
  */
 
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
-import { ShowClippedTextDirective, tip, TooltipPosition } from './show-clipped-text.directive';
+import { fireTipTransitionEndForTests, ShowClippedTextDirective, TooltipPosition } from './show-clipped-text.directive';
 import {
     ShowClippedTextDirectiveTestHelper,
     ShowClippedTextDirectiveTestHostComponent,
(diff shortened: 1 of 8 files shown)


projects/components/src/lib/directives/show-clipped-text.directive.test-helper.ts
```diff
@@ -115,7 +115,7 @@ export class ShowClippedTextDirectiveTestHelper {
         return this.tooltip.querySelector('.tooltip-content');
     }
 
-    private get tooltip(): HTMLElement {
+    public get tooltip(): HTMLElement {
```
juanmendes: Should be private

projects/components/src/lib/directives/show-clipped-text.directive.spec.ts
```diff
@@ -69,67 +65,61 @@ xdescribe('ShowClippedTextDirective', () => {
...
-        // });
+        it('changes visibility to hidden after the tooltip fades out', fakeAsync(function(this: Test): void {
+            const helper = this.clippedTextHelper;
+            helper.width = '10px';
+            helper.moveMouseOverHost();
+            // To avoid waiting a whole second
+            helper.componentInstance.directive.mouseoutDelay = 1;
+            helper.transitionTime = '1ms';
+            helper.moveMouseOffHost();
+            tick(50);
```
juanmendes: If we are calling the method ourselves, we shouldn't need the tick for 50 seconds, right? Also shouldn't need changing the transition time, right? I think `helper.transitionTime` is also not needed anywhere else so it could be deleted

PR #69, opened 2020-03-31
Changes to Datagrid + Widget Object

ryan-bradford:
# Description:
1. Improves the methods in the datagrid widget object for full testing capabilities.
2. Uses class and removed ID for the buttons
3. Featured + featured count defaults to all the buttons when undefined
4. Fixed the alignment of "..." in the button dropdown
5. Column visibility defaults to SHOWN

# Testing:
1. Updated the unit tests + used in VCD-UI

# Notes:
1. Switched to clrDgItems. When this is not used (and instead we use ngFor), selection in VCD-UI breaks. The code for the clrDgItems directive doesn't suggest that this directive enforces any client side sorting or filtering (https://github.com/vmware/clarity/blob/master/src/clr-angular/data/datagrid/datagrid-items.ts), but I'm not totally sure.

### DIFF
--- a/projects/components/src/utils/test/datagrid/datagrid.wo.ts
+++ b/projects/components/src/utils/test/datagrid/datagrid.wo.ts
@@ -34,13 +34,10 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
 
     /**
      * Gives the text content of all the cells in a row.
+     * @param rowIndex 0-based index of row
      */
     getRowValues(rowIndex: number): string[] {
-        const toReturn = [];
-        for (let columnIndex = 0; columnIndex < this.columnCount; columnIndex++) {
-            toReturn.push(this.getCellText(rowIndex, columnIndex));
-        }
-        return toReturn;
+        return this.getTexts(`${ROW_TAG}:nth-of-type(${rowIndex + 1}) ${CELL_TAG}`);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -32,6 +32,26 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
         return this.getNodeText(this.getCell(row, column));
     }
 
+    /**
+     * Gives the text content of all the cells in a row.
+     */
+    getRowValues(rowIndex: number): string[] {
+        const toReturn = [];
```
juanmendes, line 39: Couldn't we use
```typescript
return getTexts(`${ROW_TAG}:nth-of-type(${row + 1}) ${CELL_TAG}`)
```
Please document `rowIndex`, at least to specify 0 or 1 based. See

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -168,17 +209,24 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
...
+    getButtonText(buttonClass: string): string {
+        return this.getText(`button.${buttonClass}`);
+    }
+
+    /**
+     * Says if the button with the given {@param buttonClass} is disabled.
      */
-    pressTopButton(index: number): void {
-        this.click(`clr-dg-action-bar button:nth-of-type(${index + 1})`);
+    getButtonDisabled(buttonClass: string): string {
```
juanmendes, line 221: Usually, getters that return booleans are name `is` or `has` or a verb that you typically reply with true or false

`isButtonDisabled()`

Consider inverting the logic to avoid things double negatives like `!isButtonDisabled()`

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -168,17 +209,24 @@ export class ClrDatagridWidgetObject extends WidgetObject<ClrDatagrid> {
...
     /**
-     * Presses the button at the given {@param index} on the top of the grid.
+     * Gives the text of the button with the given {@param buttonClass}
+     */
+    getButtonText(buttonClass: string): string {
+        return this.getText(`button.${buttonClass}`);
+    }
+
+    /**
+     * Says if the button with the given {@param buttonClass} is disabled.
```
juanmendes, line 219: Are you sure that adding a {@param} within regular documentation is valid? I haven't heard of it

juanmendes:
Please add a comment explaining that this method  should not be public, explaining why it wad made public. Please create a JIRA issue to track it. It should explain it probably means re-working how we test the custom renderers.


    getCell(row: number, column: number): DebugElement {
        return this.findElement(`${ROW_TAG}:nth-of-type(${row + 1}) ${CELL_TAG}:nth-of-type(${column + 1})`);
    }

projects/components/src/utils/test/datagrid/datagrid.wo.ts
ryan-bradford: :( this was something that I picked up at my previous job writing backend Java D.B.C. code. But I looked it up, and JSDoc doesn't support it

ryan-bradford:
Responded to review!

PR #70, opened 2020-03-31
Address comments and add unit tests for DatagridFilter abstract class

ps37:
- Move the emission of changes and construction of form group to the abstract class
- Mke the debounce time overridable
- Add unit tests for abstract Datagrid filter class

### DIFF
--- a/projects/components/src/datagrid/filters/datagrid-filter.ts
+++ b/projects/components/src/datagrid/filters/datagrid-filter.ts
@@ -3,11 +3,11 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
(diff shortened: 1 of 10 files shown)


projects/components/src/datagrid/filters/datagrid-filter.ts
```diff
@@ -86,12 +86,31 @@ export abstract class DatagridFilter<V, C extends FilterConfig<V>> extends Subsc
      */
     changes = new Subject<null>();
 
+    ngOnInit(): void {
+        const obs = this.getDebounceTimeMs()
```
juanmendes, line 90: There should be a test for when debounce time is 0;

PR #71, opened 2020-04-02
Examples improvements + stackblitz pre-loading

ryan-bradford:
# Description:

Updates the examples application to import only the ComponentsModule rather than the individual modules. Also, makes it so the Stackblitz writer service pre-loads the template. 

# Testing

Ran the unit tests + made sure the examples app was still functional. 


juanmendes:
Sorry I forgot to mention this, can we update this so we are using the production version on  github pages?

ryan-bradford:
Production build was already enabled for GH pages, but I fixed a few errors and turned on AOT also.

PR #72, opened 2020-04-03
Namespaced all translations

ryan-bradford:
# Description

As per a conversation with Martin, added a "vcd.cc.*" namespace to all of our translations. 

# Testing

Ran "npm start" and made sure all the translations still appeared as they should. 

# Reviewer Notes:

1. I noticed in our generated i18n.json file, there is a property in every locale of "(strange character that I can't copy)": null. I checked this against VCD-UI's translations, and they also have the same strange character in their translations.ts file, so I think this is OK.

Before this change, our character used to be "(strange character that I can't copy)#": null, but with the new empty line on every file, the # goes away.


projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -1,5 +1,5 @@
 <clr-modal [clrModalOpen]="open" (clrModalOpenChange)="openChange.emit($event)" [clrModalSize]="'sm'" #modal>
-    <h3 class="modal-title">{{ dialogHeader || ('data-exporter.title' | translate) }}</h3>
+    <h3 class="modal-title">{{ dialogHeader || ('vcd.cc.data-exporter.title' | translate) }}</h3>
```
juanmendes, line 2: Can we move the defaulting to the constructor?


```ts
 /**
   * Text for the export button.
   */
  @Input() exportText: string = this.translationService.translate('vcd.cc.cancel');

```

Here and in other usages

ryan-bradford:
Added a commit that fixes the I18nModule.forChild method. Also, it changes the default logic we used to load default translations for parameters like "loading"

projects/i18n/src/lib/pipe/display-translation.pipe.ts
```diff
@@ -0,0 +1,28 @@
...
+@Injectable()
+@Pipe({
+    name: 'displayTranslation',
+    pure: false,
+})
+export class DisplayTranslationPipe implements PipeTransform {
+    constructor(private changeDetector: ChangeDetectorRef) {}
+
+    transform(value: TranslatedText): any {
+        if (typeof value === 'string') {
```
juanmendes, line 22: I think this is simple enough for a ternary
```ts
return typeof value === 'string' 
    ? value 
    : new AsyncPipe(this.changeDetector).transform(value)
```
If you don't love that, we avoid elses after return statements.

```ts
if (typeof value === 'string') {
    return value;
}
return new AsyncPipe(this.changeDetector).transform(value);
```

projects/components/src/data-exporter/data-exporter.component.ts
```diff
@@ -72,22 +73,26 @@ export class DataExporterComponent implements OnInit {
     /**
      * Text for the Dialog Header
      */
-    @Input() dialogHeader: string;
+    @Input()
+    dialogHeader: TranslatedText = this.translationService.translateAsync('vcd.cc.data-exporter.title');
```
juanmendes, line 77: We can create a mixin called `CanTranslate` that can allow our components to have a `translate()` method directly through delegation, saving typing everywhere.

See https://codesandbox.io/s/mixins-with-abstract-properties-eh9ce?file=/index.ts:123-394

projects/i18n/src/lib/pipe/translation-pipe.spec.ts
```diff
@@ -15,32 +15,6 @@ class GeneralTranslationPipe extends TranslationPipe {
...
-    checkNoChanges(): void {
-        throw new Error('Unimplemented');
-    }
-
-    reattach(): void {
-        throw new Error('Unimplemented');
-    }
-}
-
 const changeRef = {
```
juanmendes, line 18: Do we even need this? Can't we use the live one for angular?
https://angular.io/guide/testing#componentfixture-properties `changeDetectorRef `?

projects/i18n/src/lib/i18n.module.ts
```diff
@@ -69,26 +70,27 @@ export class I18nModule {
...
-                      {
-                          provide: TranslationService,
-                          useClass: MessageFormatTranslationService,
-                      },
                       {
                           provide: TranslationLoader,
+                          useClass: TranslationLoader,
                           deps: [HttpClient, extensionRoute],
                       },
+                      {
```
juanmendes, line 90: This seems like a bit of duplication

```ts
const defaultProviders = [
    {provide: TranslationService, useClass: InjectedTranslationService}
];

return !extensionRoute ? defaultProviders : [
   ...defaultProviders,
   {provide: TRANSLATIONS_COMBINED, useValue: combined},
   {provide: TranslationLoader, useClass: TranslationLoader}
];
```

PR #79, opened 2020-04-15
Fixed the subscription tracker mixin

ryan-bradford:
# Description

Updates the subscription tracker mixin to be abstract and require a ngOnDestroy method. Also, changes the subscription tracker mixin example to have an empty implementation of ngOnDestroy.

# Testing

Ran the examples app in prod mode and made sure the mixin worked.

# Future Work

Creating tslint rules:
https://jira.eng.vmware.com/browse/VDUCC-140

Signed-off-by: Ryan Bradford rbradford@vmware.com


projects/components/src/common/subscription/subscription-tracker.mixin.ts
```diff
@@ -35,8 +35,7 @@ export function SubscriptionTrackerMixin<TBase extends Type<{}>>(Base: TBase) {
             this.tracker.unsubscribeAll();
         }
 
-        ngOnDestroy(): void {
-            this.unsubscribeAll();
-        }
-    };
+        abstract ngOnDestroy(): void;
```
ps37, line 38: Why was this made abstract by removing the call to this.unsubscribeAll? Wouldn't that make all the subclasses duplicate the call to this.unsubscribeAll?

projects/components/src/datagrid/filters/datagrid-filter.ts
```diff
@@ -93,6 +93,8 @@ export abstract class DatagridFilter<V, C extends FilterConfig<V>> extends Subsc
         this.subscribe(obs, () => this.changes.next());
     }
 
+    ngOnDestroy(): void {}
```
ps37, line 96: If we go down the route of making this abstract, ain't we supposed to call unsubscribeAll here?

projects/examples/src/components/subscription/subscription-tracker-mixin.example.component.ts
```diff
@@ -18,13 +18,16 @@ import { BehaviorSubject, Observable } from 'rxjs';
...
+export class SubscriptionTrackerMixinExampleSubComponent extends SubscriptionTrackerMixin(class {})
+    implements OnInit, OnDestroy {
     @Input()
     observable: Observable<number>;
 
     ngOnInit(): void {
         this.subscribe(this.observable, value => console.log(value));
     }
+
+    ngOnDestroy(): void {}
```
ps37, line 30: Same here

juanmendes:
It seems the benefits of mixins that require returning abstract classes are made impossible by true support from AOT and TS itself.

I think the best thing to do is to stop using the SubscriptionTracker mixin. It seems that mixins outputting an abstract class is not supported because typescript loses the abstractness of the class it returns when included from a .d.ts  It works fine within a project but doesn’t work when included from node_modules.

Angular also requires that you create an ngOnDestroy on your object otherwise it won’t call  ngOnDestroy in AOT mode. It works fine for regular base classes.
With all this, I’m proposing we should not make use of mixins for abstract base classes,  e.g., subscription tracker. As much as I love the pattern 

Please remove the mixin.

image

projects/components/src/datagrid/filters/datagrid-filter.ts
ryan-bradford: The logic of calling unsubscribeAll is handled by SubscriptionTracker in the constructor. It modifies the ngOnDestroy method itself.

projects/components/src/common/subscription/subscription-tracker.mixin.ts
ryan-bradford: Removed in the most recent version, but the logic to call unsubscribeAll is in the constructor of the SubscriptionTracker

ryan-bradford:
Just added a commit to remove all mixins, rebased to origin/master, and squashed.

PR #84, opened 2020-04-22
Modify datagrid numeric filter input to have units dropdown

ps37:
This will add units dropdown next to number inputs of from and to conditionally. The condition is based on units input.

### DIFF
--- a/projects/components/src/datagrid/filters/datagrid-numeric-filter.component.html
+++ b/projects/components/src/datagrid/filters/datagrid-numeric-filter.component.html
@@ -1,15 +1,45 @@
 <div [formGroup]="formGroup" class="clr-row">
     <div class="clr-col-6">
-        <label class="clr-control-label"
-            >From:
-            <input class="from-input clr-input" type="number" [formControlName]="'from'" />
-        </label>
+        <vcd-number-with-unit-form-input
+            *ngIf="unit"
+            [label]="'vcd.cc.numeric.filter.from' | translate"
+            [formControlName]="'from'"
+            [unitOptions]="unitOptions"
+            [showUnlimitedOption]="false"
+            [min]="0"
+            [inputValueUnit]="unit"
+        >
+        </vcd-number-with-unit-form-input>
+        <vcd-form-input
+            *ngIf="!unit"
+            [type]="'number'"
+            [label]="'vcd.cc.numeric.filter.from' | translate"
+            [formControlName]="'from'"
+            [maxlength]="maxNumberLength"
+            (onEnter)="close()"
+            (onEscape)="close()"
+        ></vcd-form-input>
     </div>
 
     <div class="clr-col-6">
-        <label class="clr-control-label"
-            >To:
-            <input class="to-input clr-input" type="number" [formControlName]="'to'" />
-        </label>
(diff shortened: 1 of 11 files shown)


projects/components/src/form/form-input/form-input.component.spec.ts
```diff
@@ -12,7 +12,7 @@ import { FormInputComponent, getFormattedDateValue } from './form-input.componen
 export class VcdFormInputWidgetObject extends WidgetObject<FormInputComponent> {
     static tagName = `vcd-form-input`;
 
-    private get inputElement(): HTMLInputElement {
+    get inputElement(): HTMLInputElement {
```
juanmendes, line 15: Why are we exposing an HTML input element? We typically expose strings, booleans, number

projects/components/src/form/form-input/form-input.component.ts
```diff
@@ -71,6 +81,19 @@ export class FormInputComponent extends BaseFormControl implements AfterViewInit
      */
     @ViewChild('textInput', { static: true }) textInput: ElementRef;
 
+    /**
+     * Emitted when enter key is released for callers to handle the event like... closing a filter widget
+     * {@link DatagridNumericFilterComponent.close}
+     */
+    // tslint:disable-next-line:no-output-on-prefix
```
juanmendes, line 88: Why break ng's standards?  How about `enterClicked` and `escapeClicked`.

projects/components/src/datagrid/filters/datagrid-numeric-filter.component.html
```diff
@@ -1,15 +1,45 @@
...
+            [inputValueUnit]="unit"
+        >
+        </vcd-number-with-unit-form-input>
+        <vcd-form-input
+            *ngIf="!unit"
+            [type]="'number'"
+            [label]="'vcd.cc.numeric.filter.to' | translate"
+            [formControlName]="'to'"
+            [maxlength]="maxNumberLength"
+            (onEnter)="close()"
```
juanmendes, line 41: Why do enter and escape do the same thing? Enter should commit the change and escape should revert changes, was this a new feature you just added?

ps37: No. The functionality to close the filter widget upon pressing both Enter and Escape keys is there in Vcd UI

projects/components/src/datagrid/filters/datagrid-numeric-filter.component.spec.ts
```diff
@@ -53,24 +98,52 @@ describe('Datagrid numeric filter', () => {
...
+        describe('with units', () => {
+            beforeEach(function(this: HasFinderAndFilter): void {
+                const finderAndFilter = createDatagridFilterTestHelperWithFinder(DatagridNumericFilterComponent, {
+                    unitOptions: [...Bytes.types],
+                    unit: Bytes.MB,
+                });
+                this.finder = finderAndFilter.finder;
+                this.filter = finderAndFilter.filter;
+            });
+            // tslint:disable-next-line:max-line-length
```
juanmendes, line 131: Please avoid this `// tslint:disable-next-line:max-line-length`, break up your string if necessary

PR #85, opened 2020-04-22
Implemented lazy loading detail expansions

ryan-bradford:
# Description:

Adds the ability for a row detail to use the [clrLoading] directive to make the detail contents lazy loaded. It does this by requiring the row detail to be input as a component rather than a template, and then when it creates the component it provides the LoadingListener as a provider.

# Testing:

1. Added a unit test to test the LoadingListener provider
2. Updated the existing detail row test to work.
3. Ran the app with a lazy loaded detail row and assured it works.

detail-lazy

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.html
+++ b/projects/components/src/datagrid/datagrid.component.html
@@ -52,7 +52,7 @@
         Whereas the trackBy in ngForTrackBy is to inform Clarity on how to track changes to it's
         items. -->
     <clr-dg-row
-        *ngFor="let restItem of items; let i = index; trackBy: trackBy"
+        *ngFor="let restItem of items; let i = index; let count = count; trackBy: trackBy"
         [ngForTrackBy]="trackBy"
         [ngClass]="this.clrDatarowCssClassGetter(restItem, i)"
         [clrDgItem]="restItem"
@@ -91,9 +91,15 @@
             >
             </ng-template>
         </clr-dg-cell>
-        <ng-container ngProjectAs="clr-dg-row-detail" *ngIf="detailTemplate !== undefined">
+
+        <ng-container ngProjectAs="clr-dg-row-detail" *ngIf="detailComponent !== undefined">
             <clr-dg-row-detail *clrIfExpanded>
-                <ng-content *ngTemplateOutlet="detailTemplate; context: { record: restItem }"> </ng-content>
+                <ng-template
+                    [vcdComponentRendererOutlet]="{
+                        rendererSpec: getDetailRenderSpec(restItem, i, count)
+                    }"
+                >
+                </ng-template>
             </clr-dg-row-detail>
         </ng-container>
     </clr-dg-row>
--- a/projects/components/src/datagrid/datagrid.component.ts
+++ b/projects/components/src/datagrid/datagrid.component.ts
@@ -7,23 +7,34 @@ import {
     AfterViewInit,
     ChangeDetectorRef,
     Component,
-    ContentChild,
     ElementRef,
     EventEmitter,
     HostBinding,
+    Injector,
     Input,
     OnInit,
     Output,
-    TemplateRef,
     TrackByFunction,
+    Type,
     ViewChild,
 } from '@angular/core';
-import { ClrDatagrid, ClrDatagridFilter, ClrDatagridPagination, ClrDatagridStateInterface } from '@clr/angular';
+import {
+    ClrDatagrid,
+    ClrDatagridFilter,
+    ClrDatagridPagination,
+    ClrDatagridStateInterface,
+    LoadingListener,
+} from '@clr/angular';
 import { LazyString, TranslationService } from '@vcd/i18n';
 import { Observable } from 'rxjs';
 import { ActivityReporter } from '../common/activity-reporter';
 import { TooltipSize } from '../lib/directives/show-clipped-text.directive';
 import { DatagridFilter } from './filters/datagrid-filter';
+import {
+    ComponentRenderer,
+    ComponentRendererConstructor,
+    ComponentRendererSpec,
+} from './interfaces/component-renderer.interface';
 import {
     Button,
     ButtonConfig,
@@ -169,6 +180,24 @@ export interface PaginationConfiguration {
      * Defaults to false.
      */
     shouldShowPageNumberInput?: boolean;
+}
+
+/**
+ * The configuration object that is passed to the detail row component.
+ */
+export interface DetailRowConfig<R> {
+    /**
+     * The record that this detail row should render.
+     */
+    record: R;
+    /**
+     * The index this detail row is in the datagrid.
+     */
+    index: number;
+    /**
+     * The total number of rows in the datagrid.
+     */
+    count: number;
 }
 
 /**
@@ -378,8 +407,12 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
     TextIcon = TextIcon;
     private _columns: GridColumn<R>[];
 
-    @ContentChild(TemplateRef, { static: false }) detailTemplate!: TemplateRef<ElementRef>;
-
+    /**
+     * The component that sound be rendered for this detail row.
+     *
+     * @param R The type of record that this detail component will display.
+     */
+    @Input() detailComponent: ComponentRendererConstructor<DetailRowConfig<R>>;
     private _selectionType: GridSelectionType = GridSelectionType.None;
 
     /**
@@ -476,12 +509,10 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
     @ViewChild(ClrDatagridFilter, { static: false }) numericFilter: ClrDatagridFilter;
 
     @ViewChild(ClrDatagrid, { static: true }) datagrid: ClrDatagrid;
-
     /**
      * The pagination display within the datagrid.
      */
     @ViewChild(ClrDatagridPagination, { static: false }) paginationComponent: ClrDatagridPagination;
-
     /**
      * The activity reporter that all activites are displayed on
      */
@@ -622,6 +653,17 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
         if (response && this.actionReporter) {
             this.actionReporter.monitorGet(response);
         }
+    }
+
+    /**
+     * Gives the render spec to create the detail row for the row with the given record, at the given index, and
+     * in a datagrid with the given count of total items.
+     */
+    getDetailRenderSpec(record: R, index: number, count: number): ComponentRendererSpec<DetailRowConfig<R>> {
+        return {
+            type: this.detailComponent,
+            config: { record, index, count },
+        };
     }
 
     /**
--- a/projects/components/src/datagrid/directives/component-renderer-outlet.directive.ts
+++ b/projects/components/src/datagrid/directives/component-renderer-outlet.directive.ts
@@ -31,6 +31,11 @@ export class ComponentRendererOutletDirective<R, T> {
 
     constructor(private viewContainerRef: ViewContainerRef, private cfr: ComponentFactoryResolver) {}
 
+    /**
+     * Sets the component that this outlet should render. Where rendererSpec is the
+     * {@link ComponentRendererSpec} that details how to render the component and context
+     * is any information that the component needs to render.
+     */
     @Input()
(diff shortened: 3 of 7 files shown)


projects/components/src/datagrid/datagrid.component.html
```diff
@@ -88,9 +88,15 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
             >
             </ng-template>
         </clr-dg-cell>
-        <ng-container ngProjectAs="clr-dg-row-detail" *ngIf="detailTemplate !== undefined">
+        <ng-container ngProjectAs="clr-dg-row-detail" *ngIf="detailComponent !== undefined">
```
juanmendes, line 91: Since you're touching this, we probably don't need to to check against undefined, just `*ngIf="detailComponent"`

projects/components/src/datagrid/datagrid.component.html
```diff
@@ -88,9 +88,15 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
...
             <clr-dg-row-detail *clrIfExpanded>
-                <ng-content *ngTemplateOutlet="detailTemplate; context: { record: restItem }"> </ng-content>
+                <ng-template
+                    [vcdComponentRendererOutlet]="{
+                        rendererSpec: getDetailRenderSpec(restItem, i, count),
+                        provider: getRowLoadingListenerInjector(i)
```
juanmendes, line 96: Let's avoid the double loop here by getting the listener injector with a custom component instead. This custom component should not be exported from its module.

projects/examples/src/components/datagrid/datagrid-detail-row.example.component.ts
```diff
@@ -40,3 +47,25 @@ export class DatagridDetailRowExampleComponent {
         };
     }
 }
+
+@Component({
+    selector: 'vcd-datagrid-detail-row-sub-example',
+    template: `
+        <div [clrLoading]="loading">
+            {{ config.record }} : Record<br />
```
juanmendes, line 55: ` {{ config.record }}` will not show anything meaningful, maybe `JSON.stringify()`

projects/components/src/datagrid/datagrid.component.html
ps37: Yes. I think there should be a separate component:
- Perhaps called RowDetailComponent
- This component takes a XyzDetailComponent constructor(given as input to DataGrid) and it's context(rest item)
- This component then gets/creates the desired provider from the corresponding row around it
- And this component then passes the XyzDetailComponent constructor, it's context(rest item) and the provider extracted to componentRendererOutlet directive as:
```
{
 rendererSpec: {type: Type<XyzDetailComponent>},
 context: restItem,
 injector?: <extractedInjector>
}
```

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -172,6 +182,24 @@ export interface PaginationConfiguration {
     shouldShowPageNumberInput?: boolean;
 }
 
+/**
+ * The configuration object that is passed to the detail row component.
+ */
+export interface DetailRowConfig<R> {
```
ps37, line 188: I don't think this will be required with the above suggested approach.

If needed, we can make any XyzRowDetailComponent implement ComponentRenderer interface and pass the restItem as it's config

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -371,7 +399,7 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
     TextIcon = TextIcon;
     private _columns: GridColumn<R>[];
 
-    @ContentChild(TemplateRef, { static: false }) detailTemplate!: TemplateRef<ElementRef>;
+    @Input() detailComponent: Type<ComponentRenderer<DetailRowConfig<R>>>;
```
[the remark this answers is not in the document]
ps37: This can then just be ComponentRendererConstructor<R>

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -711,6 +741,37 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
         return this.paginationCallback(paginationData.firstItem + 1, paginationData.lastItem + 1, this.totalItems);
     }
 
+    /**
+     * Gives the render spec to create the detail row for the row with the given record, at the given index, and
+     * in a datagrid with the given count of total items.
+     */
+    getDetailRenderSpec(record: R, index: number, count: number): ComponentRendererSpec<DetailRowConfig<R>> {
```
ps37, line 748: We can get this od this method

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -711,6 +741,37 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
...
+        return {
+            type: this.detailComponent,
+            config: { record, index, count },
+        };
+    }
+
+    /**
+     * Gives the injector that will provide the {@link LoadingListener} for the row at the given index.
+     */
+    getRowLoadingListenerInjector(index: number): Injector {
```
ps37, line 758: Move this to the new component

projects/components/src/datagrid/directives/component-renderer-outlet.directive.ts
```diff
@@ -32,24 +32,28 @@ export class ComponentRendererOutletDirective<R, T> {
...
      * Attaches the passed component type to the view of this directive host
      */
-    private attachRenderer(): ComponentRef<ComponentRenderer<T>> {
+    private attachRenderer(injector?: Injector): ComponentRef<ComponentRenderer<T>> {
         if (this.componentRef) {
             this.detachRenderer();
         }
         const componentFactory = this.cfr.resolveComponentFactory(this.componentType);
-        return this.viewContainerRef.createComponent(componentFactory);
+        return this.viewContainerRef.createComponent(componentFactory, 0, injector);
```
ps37, line 56: If there is no injector defined, would this mess up the hierarchical injections this component might depend upon?

projects/components/src/datagrid/datagrid.component.ts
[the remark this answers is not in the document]
ryan-bradford: I could remove the Type, but it became ComponentRendererConstructor<DetailRowConfig<R>>

projects/components/src/datagrid/datagrid.component.html
ryan-bradford: I tried to take this approach as best as I could, let me know what you think!

projects/components/src/datagrid/datagrid.component.ts
ryan-bradford: I still needed to provide this interface to specify to the client what config object we will provide them in their details component. Because the component renderer outlet provides a single config argument this felt necessary. Let me know if you still think it can be removed, would also like to remove it of possible.

projects/components/src/datagrid/datagrid.component.ts
ryan-bradford: Because the config object is still complicated, I moved this method into the new sub component.

projects/components/src/datagrid/directives/component-renderer-outlet.directive.ts
ryan-bradford: In the new way of doing things, when we provide the LoadingListener because we are injecting it, we can assure that it is never undefined. We also know that all rows always provide LoadingListener. Does this sound correct?

ryan-bradford:
Just responded to the review comments y'all left. Mainly, created the new details component which simplified the creation of the details.

projects/components/src/datagrid/details/detail-wrapper.component.ts
```diff
@@ -0,0 +1,64 @@
...
+     * The component that sound be rendered for this detail row.
+     *
+     * @param R The type of record that this detail component will display.
+     */
+    @Input() detailComponent: ComponentRendererConstructor<DetailRowConfig<R>>;
+
+    /**
+     * Constructs this detail wrapper with the injected loading listener.
+     */
+    constructor(public loadingListener: LoadingListener) {
```
juanmendes, line 43: If we instead do

```ts
constructor(public injector:Injector) { }
```

Would that allow the custom detail component to access the rest manager in vcd_ui?

Right now it looks like the custom detail component wouldn't have access to anything it could need through DI

projects/components/src/datagrid/details/detail-wrapper.component.html
```diff
@@ -0,0 +1,9 @@
+<clr-dg-row-detail *clrIfExpanded>
+    <ng-template
+        [vcdComponentRendererOutlet]="{
+            rendererSpec: getDetailRenderSpec(),
+            provider: injector
```
juanmendes, line 5: shouldn't this say:

`injector: injector`

according to https://github.com/vmware/vmware-cloud-director-ui-components/pull/85/files/672bccf0404e7a2ac7d6abd65363b0c006eaf8e1..70a567ee70032312b15d55c4544639b5d5d549ba#diff-7650ed1cc58e90f46bbe8a7229717fbcR39 ?

projects/components/src/datagrid/directives/component-renderer-outlet.directive.ts
```diff
@@ -32,24 +32,28 @@ export class ComponentRendererOutletDirective<R, T> {
     constructor(private viewContainerRef: ViewContainerRef, private cfr: ComponentFactoryResolver) {}
 
     @Input()
```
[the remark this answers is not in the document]
juanmendes: Sorry, I meant add doc to `set vcdComponentRendererOutlet`. That's what callers have to be educated on, constructors of components that are constructed with DI don't need documentation. Feel free to remove the comment on the constructor(s)

projects/components/src/datagrid/details/detail-wrapper.component.html
ryan-bradford: Ok, so when I looked at this comment I was confused because why was this working in the first place if I had such a fundamental thing wrong? So, I removed all references to new Injectors and Providers and as it turns out, when you create a component dynamically the dependency chain is not broken, it's only when you create a component within some dynamic template. So we can remove all references to Injection and Provider and this whole detail wrapping component all together

projects/examples/src/components/datagrid/datagrid-detail-row.example.component.ts
```diff
@@ -14,10 +14,27 @@ interface Data {
  * Shows an expandable detail template within the rows.
  */
 @Component({
-    selector: 'vcd-datagrid-show-hide-example',
+    selector: 'vcd-datagrid-detail-row-example',
     template: `
-        <vcd-datagrid [gridData]="gridData" (gridRefresh)="refresh($event)" [columns]="columns">
-            <ng-template let-record="record"> DETAILS: {{ record.value }} </ng-template>
+        <div>
```
juanmendes, line 19: A `div` should be used for containers that do layout. If it's just a paragraph, use `<p>`

projects/examples/src/components/datagrid/datagrid-detail-row.example.component.ts
```diff
@@ -14,10 +14,27 @@ interface Data {
...
+            The datagrid allows the client to pass in an arbitrary component to be rendered as the expandable detail
+            content of the rows. This is passed in as [detailComponent] and has support for using [clrLoading] to make
```
juanmendes, line 21: Mark code with `<code></code>`

projects/examples/src/components/datagrid/datagrid-detail-row.example.component.ts
```diff
@@ -40,3 +60,42 @@ export class DatagridDetailRowExampleComponent {
...
+            this.loading = false;
+        }, 2000);
+    }
+}
+
+@Component({
+    selector: 'vcd-datagrid-detail-row-sub-example',
+    template: `
+        <div>
+            {{ JSON.stringify(config.record) }} : Record<br />
```
juanmendes, line 91: This seems backwards, I usually think key, value, key, value, this seems to be value, key

A simple HTML tag for key values is `<dl> <dt> <dd>` See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

PR #86, opened 2020-04-23
Add wild card at both start and end of a string filter input

ps37:
This will add on option to pad * at both start and end of a string filter input

### DIFF
--- a/projects/components/src/datagrid/filters/datagrid-string-filter.component.spec.ts
+++ b/projects/components/src/datagrid/filters/datagrid-string-filter.component.spec.ts
@@ -65,7 +65,7 @@ describe('Datagrid string filter', () => {
         it('returns a FIQL string with equalTo operator and * added at both the beginning and at the end of a filter input', function(this: HasDgStringFilter): void {
             this.filter.config = {
                 ...this.filter.config,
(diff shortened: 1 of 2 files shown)


projects/components/src/datagrid/filters/datagrid-string-filter.component.spec.ts
```diff
@@ -61,5 +61,14 @@ describe('Datagrid string filter', () => {
             this.filter.setValue('test-input');
             expect(this.filter.getValue()).toEqual(`${queryFieldName}==*test-input`);
         });
+        // tslint:disable-next-line:max-line-length
+        it('returns a FIQL string with equalTo operator and * added at both the beginning and at the end of a filter input', function(this: HasDgStringFilter): void {
```
juanmendes, line 65: There's no reason to disable max-line-length here,  it makes it harder to read the line. Please break up the string with `+` or put parameters into separate lines, or both and let prettier do its job. 

Simplification to keep it on one line, even if it didn't fit, you could just break it up + into different lines so we can still read without scrolling or random wrapping

```ts
it(
    'returns FIQL using the equalTo operator, ' +
    'wrapping the user input with *s',
     function(this: HasDgStringFilter): void {
         // I code here
```

PR #88, opened 2020-04-23
Adds trackBy to datagrid rows ngFor

ryan-bradford:
# Description:

Adds back in a trackBy clause to ngFor and fixes the related selection bugs that it causes.

# Details:

There is a known and open bug in Clarity that using trackBy and refreshing the data breaks the select all functionality (https://github.com/vmware/clarity/issues/2265). This is due to orphaned selected items/rows in the datagrid. 

The fix to the problem is either to remove trackBy (which we tried and cannot do), use clrDgItems (which we tried and cannot do), or follow the solution in this Stackblitz linked from the above GitHub issue (https://stackblitz.com/edit/issue-2265-workaround?file=src%2Fapp%2Fapp.component.ts). This solution involves adding a detectChanges when you update the grid data and manually calling grid.rows.notifyOnChanges. Not totally ideal, but it seems to be the only way. 

# Testing:

1. Used the grid selection example and made sure selection worked for both multi and single selection.
2. Ran the example that shows a grid where pressing a button opens a modal and also refreshes the data. Showed that the modal remains open.
3. Updated the expandable row example to refresh the row on sorting and showed that the detail row stays open.
4. Used this build in the IPsec page and assured that it fixes the details-close-on-polling bug.

# Selection Gif

As you can see in the gif, when the grid data is refreshed, the clr-dg-row is not recreated and the selection maintains as it should.

grid-selection

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.html
+++ b/projects/components/src/datagrid/datagrid.component.html
@@ -48,6 +48,9 @@
         </clr-dg-filter>
     </clr-dg-column>
 
+    <!--The trackBy in the ngFor is to inform Angular on how to track changes to clrDgRow.
+        Whereas the trackBy in ngForTrackBy is to inform Clarity on how to track changes to it's
+        items. -->
     <clr-dg-row
         *ngFor="let restItem of items; let i = index; trackBy: trackBy"
         [ngForTrackBy]="trackBy"
--- a/projects/components/src/datagrid/datagrid.component.html
+++ b/projects/components/src/datagrid/datagrid.component.html
@@ -49,7 +49,7 @@
     </clr-dg-column>
 
     <clr-dg-row
-        *ngFor="let restItem of items; let i = index"
+        *ngFor="let restItem of items; let i = index; trackBy: trackBy"
         [ngForTrackBy]="trackBy"
         [ngClass]="this.clrDatarowCssClassGetter(restItem, i)"
         [clrDgItem]="restItem"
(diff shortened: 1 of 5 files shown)


projects/components/src/datagrid/datagrid.component.html
```diff
@@ -49,7 +49,7 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
     </clr-dg-column>
 
     <clr-dg-row
-        *ngFor="let restItem of items; let i = index"
+        *ngFor="let restItem of items; let i = index; trackBy: trackBy"
```
juanmendes, line 52: Can we add a comment here? It's unclear why we need both trackBy and `ngForTrackBy`

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -627,28 +635,35 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
...
             if (this.datagrid.selection.current) {
-                this.datagrid.selection.current = this.datagrid.selection.current.filter((selected, selectedIndex) => {
-                    const found = this.items.find(
-                        (item, itemIndex) => this.trackBy(itemIndex, item) === this.trackBy(selectedIndex, selected)
-                    );
-                    return found;
-                });
+                const current = [...this.datagrid.selection.current];
+                this.datagrid.selection.clearSelection();
+                const nextSelection = current
```
juanmendes, line 653: Add a comment explaining that we are mapping from the old records to the new records since they will typically new references.

It may be worth an appropriately named private method

```ts

/**
 * Given an existing selection that is made up of records that are about to become stale, and new records to be loaded into the grid, returns the selection mapped to records from the newly added records, excluding any records that may not be present in the new selection because they are not present in the new data.
 */
mapSelectedRecords(currentSelection: T[], newRecords: T[]): T[] {

}
```

PR #90, opened 2020-04-24
Add disabled functionality to number with form input control value accessor

ps37:
Disabling the formControl of number-with-form input component should disable the underlying input and select t form controls.

This functionality was missing before. Found this while integrating the component in vcd-ui.


ps37:
> Bump components version in package.json?

I will do that once I get a ship it from both of you and ready to push to the main repo

PR #91, opened 2020-04-27
Fixed data exporter injection issue

ryan-bradford:
# Description:

Resolves a code injection issue that was present in our data exporter. Pretty much, if the user gives data of which any given cell begins with the characters '+', '-', '@', or '=', the user is prompted if she wants to sanitize her data. If she chooses yes, each cell is prefixed with the tab character. If she chooses no, the data is given in its original form. 

# Testing:

1. Added unit tests to both the data exporter and the CSV service.
2. Updated the example so it shows the injection feature. Tested that both the yes and no buttons function as designed. Gif seen below:

# Reviewer Notes:
1. Some of the method names I was unsure about. I wasn't sure if I should use "purge", "purify", "sanitize", or some other verb to describe this action. 
2. The logic I use to handle displaying the case where the data was unsafe felt strange. Caching both "_data" and "_shouldPurifyData" felt weird and not very clean. Open to other ways to handle this. 

data-exporter


projects/components/src/assets/resources/en.properties
```diff
@@ -10,6 +10,9 @@ vcd.cc.data-exporter.title=Data Exporter
 vcd.cc.loading=Loading
 vcd.cc.grid.default.pagination={firstItem} - {lastItem} of {totalItems} row(s)
 vcd.cc.unlimited=Unlimited
+vcd.cc.yes=Yes
+vcd.cc.no=No
+vcd.cc.bad.data=The data you have exported contains potential code injection. Would you like to sanitize this data to prevent code injection?
```
juanmendes, line 15: How about

The data being exported contains information that would cause it to execute formulas if opened with a spreadsheet. Would you like to sanitize this data to prevent code injection?

projects/components/src/data-exporter/csv-exporter.service.ts
```diff
@@ -16,8 +16,17 @@ export class CsvExporterService {
...
      * @param rows 2D array of data. First row is the names for the fields
      */
-    public createCsv(rows: any[][]): string {
-        return rows.map(row => processRow(row)).join('\n');
+    public createCsv(rows: any[][], shouldPurify = false): string {
+        return rows.map(row => processRow(row, shouldPurify)).join('\n');
+    }
+
+    /**
+     * Says if the CSV that would result from the export is at risk of code injection
```
juanmendes, line 24: "Whether" sounds better than "Says"

projects/components/src/data-exporter/csv-exporter.service.ts
```diff
@@ -16,8 +16,17 @@ export class CsvExporterService {
...
+     * @param rows 2D array of data. First row is the names for the fields
+     * @param shouldPurify if the tab character should be put in front of any formula cells.
```
juanmendes, line 26: First letter of documentation should be capital

projects/components/src/data-exporter/csv-exporter.service.spec.ts
```diff
@@ -40,6 +55,13 @@ describe('CsvExporterService', () => {
...
             expect(service.createCsv([['a', 'b'], [null, undefined]])).toEqual('a,b\n,');
         });
+
+        it('adds a tab character if the value contains any special characters', () => {
+            const service: CsvExporterService = TestBed.get(CsvExporterService);
+            expect(service.createCsv([['+', '-', '@', '=', 'a+'], [null, undefined]], true)).toEqual(
+                '\t+,\t-,\t@,\t=,a+\n,'
+            );
+        });
     });
```
juanmendes, line 65: Can we split this into a separate test making sure it doesn't sanitize if control characters are not at the beginning of string. Let's always try to have separate tests for different conditions instead of testing two different things in the same `it()`. The rule is that the text of the it should match the assertion, here there's no it explaining that it does not attempt to sanitize a  control character that is not the first one, but you do seem to be asserting it.

projects/components/src/data-exporter/csv-exporter.service.ts
```diff
@@ -44,19 +53,28 @@ export class CsvExporterService {
     }
 }
 
+const SPECIAL_CHAR = new RegExp('^\\+|^-|^=|^@');
```
juanmendes, line 56: Use literals instead of constructors whenever possible, that way you don't need to double escape.

This could be:

```ts
// Dash has to be the first one in the character sequence, otherwise it would be treated as a range operator. 
// If you think this is too tricky, you could escape it as \-
const LEADING_CONTROL_CHAR = /^[-+=@]/;
```

ryan-bradford:
Fixed the comments posted by Juan!

projects/components/src/data-exporter/data-exporter.component.ts
```diff
@@ -131,8 +131,8 @@ export class DataExporterComponent implements OnInit {
 
     private _open = false;
 
-    _shouldPurifyData = false;
-    _data: any[][];
+    shouldSanitizeData = false;
```
juanmendes, line 134: This seems to really mean  `dataHasPotentialInjection` since we should only sanitize the data if the user chooses Yes.

projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -44,10 +44,10 @@ <h3 class="modal-title">{{ dialogHeader | lazyString }}</h3>
         </ng-container>
         <ng-template #purifyDataButtons>
             <div class="modal-footer">
-                <button type="button" class="btn btn-outline no" (click)="_downloadData(_data, false)">
+                <button type="button" class="btn btn-outline no" (click)="downloadData(data, false)">
                     {{ noText | lazyString }}
                 </button>
-                <button type="button" class="btn btn-primary yes" (click)="_downloadData(_data, true)">
+                <button type="button" class="btn btn-primary yes" (click)="downloadData(data, true)">
```
juanmendes, line 50: It looks like we don't really need to pass data here if we're already storing it as state? Then we could make `data` private?

PR #92, opened 2020-05-06
Add API to modify grid columns list by adding/removing/updating a column

ps37:
Notify listeners about columns updated event
Add methods to add or remove a column from list grid columns

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { Component, HostBinding, InjectionToken, Type, ViewChild } from '@angular/core';
+import { Component, HostBinding, ViewChild } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { LoadingListener } from '@clr/angular';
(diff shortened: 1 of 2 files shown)


projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -520,6 +535,47 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
...
+            return;
+        }
+        const colIndex = this.findColumnIndex(col);
+        if (colIndex !== null) {
+            this._columns.splice(colIndex, 1);
+            this.updateColumnsConfig();
+        }
+    }
+
+    private findColumnIndex(col: GridColumn<R>): number {
```
juanmendes, line 570: Why return null instead of -1? Most `findIndex` methods return -1 when an index is not found

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -388,6 +391,19 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
         }
         return [];
     }
+    /**
+     * Stream to send push notifications when ever columns are updated by the calling components using this data grid
+     */
+    private columnsUpdated = new Subject<boolean>();
+
+    columnsUpdatedObservable = this.columnsUpdated.asObservable();
```
juanmendes, line 399: Is this supposed to be `@Output`? You documented a private `Subject` but didn't document the public API? Only use observables if you absolutely must. `EventEmitters` are the standard way to communicate between components, you should always use that when possible.

What is the boolean value passed to the events for?

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -255,12 +255,15 @@ interface ColumnConfigInternal<R, T> extends GridColumn<R> {
...
-     * Sets the configuration of columns on the grid and updates the {@link columnsConfig} array
+     * Sets the configuration of columns on the grid and updates the {@link columnsConfig} array. Also pushes
+     * notifications for listeners to make changes to the _columns array
      */
     @Input()
     set columns(cols: GridColumn<R>[]) {
         this._columns = cols;
-        this.getColumnsConfig();
+        this.updateColumnsConfig();
+        // Push notification
```
juanmendes, line 265: Comment doesn't tell me anything the code doesn't already

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -884,6 +884,88 @@ describe('DatagridComponent', () => {
...
+                this.finder.hostComponent.grid.addColumn({
+                    displayName: 'Column2',
+                    renderer: 'name2',
+                });
+                this.finder.detectChanges();
+                expect(this.finder.hostComponent.grid.columns.length).toBe(2);
+                expect(this.finder.hostComponent.grid.columns[1].displayName).toBe('Column2');
+            });
+            it(
+                'updates a existing column if a column exists with same display name as the column passed ' + 'in',
```
juanmendes, line 924: Why are you concatenating this if it fits in one line?

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -884,6 +884,88 @@ describe('DatagridComponent', () => {
                 expect(this.clrGridWidget.component.columns.first.customFilter).toEqual(true);
             });
         });
+
```
juanmendes, line 887: Please add unit test for the `columnsChanged` event. I would expect some test showing the difference about an event being fired when you set columns but not fired when you add/remove columns through the special methods?

projects/components/src/datagrid/datagrid.component.spec.ts
ps37: If I don't break the string, The Prettier Git hook moves the `function(this: HasFinderAndGrid): void {` above and that exceeds the maximum line length

PR #95, opened 2020-05-11
Implemented data exporter rework

ryan-bradford:
# Description:

Reworks the data exporter to have the design as created in [this ticket](https://jira.eng.vmware.com/browse/VDUCC-167). Summary of the features:

1. Only one screen
2. Show a status indicator that says the type of action going on (I.E. downloading, sanitizing, other)
3. should have three checkboxes
    1. Should Sanitize (has an info icon)
    2. Show user friendly names (has an info icon)
    3. Select All columns
         - Unchecking this shows a list of all columns with checkboxes on the right
         - Has a fixed number of columns and the rest overflow to some container
4. Has a fixed size and width and buttons don't move when things are checked

# Testing:

Created unit tests to test the new features and tested using the examples app to make sure it works. 

# Reviewer Notes:
1. Some of the wordings of the info messages/hints I was unsure about. 
2. Should we break off the multi-select dropdown HTML into it's own component?

# Gif:

data-exporter


projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -1,56 +1,82 @@
...
+                    <i>{{ friendlyNamesHint | lazyString }}</i>
+                    <clr-checkbox-wrapper>
+                        <input class="sanitize-cells" type="checkbox" clrCheckbox [formControlName]="'sanitize'" />
+                        <label style="display: flex; flex-direction: row;">
+                            {{ sanitizeCheckboxLabel | lazyString }}
+                            <div style="flex: 1"></div>
+                            <clr-tooltip>
+                                <clr-icon clrTooltipTrigger shape="info-circle" size="24"></clr-icon>
+                                <clr-tooltip-content clrPosition="bottom-left" clrSize="md" *clrIfOpen>
+                                    <span>{{ sanitizeInfoMessage | lazyString }}</span>
```
ryan-bradford, line 29: Change these and above to be using clr-signpost

projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -1,56 +1,108 @@
...
+                                type="checkbox"
+                                clrCheckbox
+                                [formControlName]="'selectAll'"
+                                (click)="onClickCheckAll($event)"
+                            />
+                            <label>{{ exportAllText | lazyString }}</label>
+                        </clr-checkbox-wrapper>
+                        <clr-dropdown *ngIf="!isSelectAllEnabled">
+                            <clr-icon
+                                shape="caret down"
```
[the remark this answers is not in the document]
ryan-bradford: Also made it so that it hides the bubbles when you open the dropdown, like we talked about.

projects/components/src/data-exporter/data-exporter.component.ts
```diff
@@ -83,10 +83,16 @@ export class DataExporterComponent implements OnInit {
     cancelText: LazyString = this.translationService.translateAsync('vcd.cc.cancel');
```
juanmendes, line 83: Please update header year to say 2019-2020 in any files that were touched

projects/components/src/data-exporter/data-exporter.component.ts
```diff
@@ -163,28 +203,44 @@ export class DataExporterComponent implements OnInit {
...
-            selectedColumns: this.columns.filter(col => this.formGroup.controls[col.fieldName].value),
+            selectedColumns: this.selectedColumns,
         });
     }
 
-    onClickCheckAll(): void {
-        for (const column of this.columns) {
-            this.formGroup.controls[column.fieldName].setValue(true);
+    onClickCheckAll(value: boolean): void {
+        if (!this.optionsFormGroup.get('selectAll').value) {
```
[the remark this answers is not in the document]
juanmendes: We also try to avoid using hard code strings to refer to controls and this seems to be the first time we need to hard code access to controls, yes we may not need it but doing the following would help other pieces of the code also.

```ts
// Create consts for the field names, they will only be used for initialization and in the control getter
const UiFields = createStringEnum([
    "field1",
    "field2",
    ...
]);

// Then we can use UiFields.field1  with some type safety (edited) 
// If bringing in createStringEnum is too much, we can just use
enum Fields = {
  field1: 'field1',
  field2: 'field2'
}
```
## HTML initialization
If you change the three places where you are returning the `formcontrol.value` such as [get isSelectAllEnabled()](https://github.com/vmware/vmware-cloud-director-ui-components/pull/95/files?file-filters%5B%5D=.json&file-filters%5B%5D=.properties&file-filters%5B%5D=.ts#diff-8ec3d6050a1425538628003d98cfe85dR234) to return the form control, they can also be used  to initialize the `[formControl]` in the HTML instead of using `[formControlName]`

## Testing considerations
You can now easily using the form control without having to write a wrapper for the checkbox since it's part of the component's API.

`comp.showFriendlyNamesControl.setValue(true)`
`expect(comp.showFriendlyNamesControl.value).toBe(false)`

projects/components/src/assets/resources/en.properties
```diff
@@ -12,7 +12,15 @@ vcd.cc.grid.default.pagination={firstItem} - {lastItem} of {totalItems} row(s)
...
-vcd.cc.bad.data=The data being exported contains information that would cause it to execute formulas if opened with a spreadsheet. Would you like to sanitize this data to prevent code injection?
+vcd.cc.sanitize.question=Sanitize Data
+vcd.cc.sanitize.info=The data being exported contains information that would cause it to execute formulas if opened with a spreadsheet. Would you like to sanitize this data to prevent code injection?\nhttps://payatu.com/csv-injection-basic-to-exploit
+vcd.cc.sanitize.hint=prefix cells with tab to prevent injection?
+vcd.cc.friendly.names.question=Friendly Field Names
+vcd.cc.friendly.names.info=If the headers of the output CSV sheet should be the translated, friendly field names or the raw field names from the object.
+vcd.cc.friendly.names.hint=user friendly column headers or object field names?
+vcd.cc.export.all=Export All Columns
+vcd.cc.exporter.downloading=Downloading...
+vcd.cc.exporter.sanitizing=Sanitizing...
```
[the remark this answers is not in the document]
ryan-bradford: Tested this out, and if I add a setTimeout, the Sanitizing message does appear. Which is preferable? Not showing the message or adding the setTimeout?

projects/components/src/data-exporter/data-exporter.component.ts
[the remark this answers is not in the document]
ryan-bradford: Changed to used the enum and getters to manage the form controls. 

And as far as the method onClickCheckAll goes, I gave the input type wrong in the code. The actual input type is a MouseEvent, which didn't seem to directly contain the information I needed. I removed the parameter to this method.

projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -1,56 +1,109 @@
...
+                                class="export-all"
+                                type="checkbox"
+                                clrCheckbox
+                                [formControl]="this.selectAllControl"
+                            />
+                            <label>{{ exportAllText | lazyString }}</label>
+                        </clr-checkbox-wrapper>
+                        <clr-dropdown *ngIf="!this.selectAllControl.value">
+                            <clr-icon
+                                shape="caret down"
```
juanmendes, line 57: Delete this line?

projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -1,56 +1,109 @@
...
+                                </ul>
+                            </clr-dropdown-menu>
+                        </clr-dropdown>
+                    </div>
                 </li>
             </ul>
-            <div class="progress" [ngClass]="{ loop: progress == -1, pending: isRequestPending }">
-                <progress max="100" value="{{ progress * 100 }}"></progress>
+            <div
+                [ngStyle]="shouldShowBubbles ? { visibility: 'visible' } : { visibility: 'hidden' }"
```
juanmendes, line 84: ```{ visibility: shouldShowBubbles ? 'visible' : 'hidden' }```

projects/components/src/data-exporter/data-exporter.component.scss
```diff
@@ -1,11 +1,71 @@
...
+    display: block;
+}
+
 button.select-all {
     margin: 0;
     padding: 0;
 }
+
+clr-tooltip-content {
+    width: 200px !important;
```
juanmendes, line 49: Important must have an explanation, it's a last resort. Can't you just add specificity? Where is this used?

projects/components/src/data-exporter/data-exporter.component.scss
```diff
@@ -1,11 +1,71 @@
...
+}
+
+.column-container {
```
juanmendes, line 52: Why do we need to fuss with the CSS? Can't we just use Clarity styles like 

```<span class="badge badge-info">2</span>```?

File: projects/components/src/data-exporter/data-exporter.component.scss (hunk 8 above)
Comment (ryan-bradford) [reply projects/components/src/data-exporter/data-exporter.component.scss:49]: That was outdated code, removed! Thanks!

File: projects/components/src/data-exporter/data-exporter.component.scss (hunk 9 above)
Comment (ryan-bradford) [reply projects/components/src/data-exporter/data-exporter.component.scss:52]: This is custom style code copied from VCD-UI that makes the bubbles + X look like they look in the VMs screen. I don't know if there are any Clarity classes that apply this specific styling.

**Comment by ryan-bradford** [comment]
Updated per review!

File: projects/components/src/data-exporter/data-exporter.component.scss [hunk 10]
```diff
@@ -5,6 +5,35 @@ div.progress {
...
+    flex: 1;
+}
+
+.column-dropdown-menu {
+    height: 200px;
+    overflow: scroll;
+}
+
+.dropdown-button {
+    vertical-align: middle;
```
juanmendes, line 24: Can we nudge this up by a pixel? I don't think we need `vertical-align`

PR #100, opened 2020-05-26
Fixed link to non primitive types in the examples container

ryan-bradford:
# Description

* Types did not appear where we had a default value or used a setter.
* Added links to Compodoc for complex types in the types column.

Relevant Live Docs Review: https://github.com/vmware/live-docs/pull/5
Relevant Compodoc Review: https://github.com/ryan-bradford/compodoc/pull/1

# Changes
1. Certain inputs whose type can be inferred from the default value need an explicit type now. These are the ones where the default value isn't either a Number, String, or Boolean or a `new ...()`. So, in this case Number.MAX needs a type value.
2. Functions only appear nice in Compodoc if they are written as interfaces. So, we disabled the callable-types rule and changed function types to be interfaces. 
3. Fixed a few places where we missed giving a type or a default value. 

# Testing

Ran the examples app with the updated version of live-docs. 
1. On the datagrid page, the type of the setter buttonConfig appears as ButtonConfig<R> and when clicking the link it brings you to http://localhost:4200/compodoc/interfaces/ButtonConfig.html. 
2. On the data exporter page, the type of cancelText appears as LazyString and clicking brings you to http://localhost:4200/compodoc/miscellaneous/typealiases.html#LazyString
3. On the datagrid page, the type of  the columns setter appears as GridColumn<R>[] and clicking links to http://localhost:4200/compodoc/interfaces/GridColumn.html
4. On the form select page, showAsterisk appears as a boolean even though it uses a default value. 
5. On the form input page, hintPosition appears as a string even though it uses a default value. 


juanmendes:
Can we please add to the description what we had to do to the code to get it working?

* Must set type of some ??? types of fields
* Disabled lint rule x because of y

ryan-bradford:
Updated the description with what you asked for

ryan-bradford:
Updated the version of live-docs, this is good to merge when aporoved.

PR #101, opened 2020-05-26
Fixed problems with the datagrid

ryan-bradford:
# Description

Fixes some reported problems with the datagrid. Namely:

1. Emit an event when selection changed.
2. contextualButtonConfig should be optional
3. Set the height of a column

# Testing:

1. Added tests for selection changed
2. Added tests for width being set.
3. Added tests for optional contextualButtonConfig

# Widths Screenshot

image

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { Component, HostBinding, ViewChild } from '@angular/core';
+import { Component, EventEmitter, HostBinding, ViewChild } from '@angular/core';
 import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { LoadingListener } from '@clr/angular';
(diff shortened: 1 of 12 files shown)


projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
```diff
@@ -253,6 +253,11 @@ export interface GridColumn<R> {
      * Whether to show the column as sortable. Defaults to true
      */
     sortable?: boolean;
+
+    /**
+     * The width of this single column in pixels.
+     */
+    width?: number;
```
juanmendes, line 260: This should not be hardcoded to pixels, people may want a different unit such as em, en, percentage... https://developer.mozilla.org/en-US/docs/Web/CSS/length It should be a string

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -173,6 +173,26 @@ describe('DatagridComponent', () => {
...
+                    this.finder.detectChanges();
+                });
+
+                it('sets the width of the column to the given height', function(this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.columns = [
+                        { displayName: 'Name', renderer: 'name', width: 100 },
+                        { displayName: 'Name', renderer: 'name' },
+                    ];
+                    this.finder.detectChanges();
+                    expect(this.clrGridWidget.getColumnWidth(2)).toEqual('100px');
```
juanmendes, line 191: I don't follow the indexes, there are two columns but we get the third and fourth columns?

ryan-bradford: I believe this is a problem with the pollution of Debug Elements in the Datagrid. This is fixed in Clarity 3 and I'll add a comment as such.

ryan-bradford:
Responded to the review, changed the button type to be link, and fixed how we set the column width. There might be a better solution (rip setTimeout), so totally open to ideas!

projects/components/src/datagrid/datagrid.component.spec.ts
ryan-bradford: The test started failing more sporadically, so I ended up commenting it out.

projects/examples/src/components/datagrid/datagrid-three-renderers.example.component.scss
```diff
@@ -0,0 +1,16 @@
+/*!
```
juanmendes, line 1: We should not be mixing examples, can we have an example for how to set column width? Can we show some `percentages`, some using `em`, some using `pt`, and probably `px` too.

Right now there's nowhere in the app that would lead me to find this feature.

juanmendes: We should embedding values in a selector name. It's bad practice, I know in this case it doesn't matter because there's no real semantic meaning, but it sets a bad example.

`col-a, col-b, col-c`. You can probably show how you can do it without any TS, just using `nth-of-type`.

Make sure the example text narrates what it's trying to show.

PR #113, opened 2020-06-09
Interface for the action menu component

ps37:
ActionItem interface that defines the type of actions that can be given as input to the action menu component.

The component itself with inputs and methods needed for plumbing the model required for rendering the view.

### DIFF
--- a/projects/components/src/action-menu/action-menu.component.spec.ts
+++ b/projects/components/src/action-menu/action-menu.component.spec.ts
@@ -34,7 +34,7 @@ describe('ActionMenuComponent', () => {
 
     describe('set actions', () => {
         it('marks the actions with no actionType as ActionType.CONTEXTUAL', () => {
-            testHostComponent.actions = ACTIONS_WITH_NO_ACTION_TYPES;
+            testHostComponent.actions = getNewArray(ACTIONS_WITH_NO_ACTION_TYPES);
             fixture.detectChanges();
             testHostComponent.actionMenuComp.actions.forEach(action => {
                 expect(action.actionType).toEqual(ActionType.CONTEXTUAL);
@@ -50,7 +50,7 @@ describe('ActionMenuComponent', () => {
             expect(actionDisplayConfig.staticActionStyling).toEqual(ActionStyling.INLINE);
         });
         it('sets display config options to given values when a input is given', () => {
-            testHostComponent.actionDisplayConfig = ACTION_DISPLAY_CONFIG;
+            testHostComponent.actionDisplayConfig = getNewObj(ACTION_DISPLAY_CONFIG);
             fixture.detectChanges();
             const actionDisplayConfig = testHostComponent.actionMenuComp.actionDisplayConfig;
             expect(actionDisplayConfig.contextual.featuredCount).toEqual(2);
@@ -62,7 +62,7 @@ describe('ActionMenuComponent', () => {
             'sets shouldShowIcon to true, shouldShowText to false and shouldShowTooltip to true when buttonContents ' +
                 'is set to icon',
             () => {
-                testHostComponent.actionDisplayConfig = ACTION_DISPLAY_CONFIG;
+                testHostComponent.actionDisplayConfig = getNewObj(ACTION_DISPLAY_CONFIG);
                 fixture.detectChanges();
                 expect(testHostComponent.actionMenuComp.shouldShowIcon).toBeTruthy();
                 expect(testHostComponent.actionMenuComp.shouldShowText).toBeFalsy();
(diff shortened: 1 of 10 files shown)


projects/components/src/action-menu/action-menu.component.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+})
+class TestHostComponent<R extends Record> {
+    @ViewChild(ActionMenuComponent, { static: true }) actionMenuComp: ActionMenuComponent<R, HandlerData>;
+
+    actionDisplayConfig: ActionDisplayConfig;
+
+    actions: ActionItem<R, HandlerData>[] = [];
+}
+
+const ACTION_DISPLAY_CONFIG = {
```
juanmendes, line 257: Do not re-use objects , create a function that returns a new object so that every caller gets a separate instance. Change all instances in change

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -0,0 +1,194 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { Component, Input } from '@angular/core';
+import { ActionDisplayConfig, ActionItem, ActionStyling, ActionType, TextIcon } from '../common/interfaces';
+
+/**
```
juanmendes, line 9: What do T and R stand for? Please always document your generic types

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Used for disabling the menu bar or menu dropdown
+     */
+    @Input() disabled = false;
+
+    /**
+     * A switch to hide static actions. For example, Used when we want to display only contextual actions per row or
+     * card
+     */
+    @Input() showStaticActions = true;
```
juanmendes, line 35: Please add type information, compodoc doesn't infer it.

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -0,0 +1,194 @@
...
+     */
+    private _actionDisplayConfig: ActionDisplayConfig = {
+        contextual: {
+            featuredCount: 0,
+            styling: ActionStyling.INLINE,
+            buttonContents: TextIcon.TEXT,
+        },
+        staticActionStyling: ActionStyling.INLINE,
+    };
+    @Input() set actionDisplayConfig(config: ActionDisplayConfig) {
```
juanmendes, line 75: Please move documentation from the backing private to here

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -0,0 +1,194 @@
...
+        if (!config) {
```
juanmendes, line 76: Why is this null check required? Should it default to something when null is passed in? Please document this behavior

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -0,0 +1,194 @@
...
+        );
+        return this.getAvailableActions(contextualActions, selection);
+    }
+
+    /**
+     * Execute the {@link ActionItem.handler} function by passing it {@link #selectedEntities} and
+     * {@link ActionItem.handlerData} as arguments
+     */
+    runActionHandler(action: ActionItem<R, T>): void {
+        const actionHandlerData = [];
```
juanmendes, line 171: Should just call

```ts
action.handler(this.selectedEntities, action.handlerData1)
```

RIght now you're breaking the signature if there are no selected entities

projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/**
+ * List of different type of action buckets
+ */
+export enum ActionType {
+    /**
+     * Global actions that are displayed always irrespective of the context. These get shown as the first set of actions
```
juanmendes, line 11: Replace `These get shown` with `These display` throughout

projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Type of {@link ActionItem.handler} function
+ */
+export type ActionHandlerType<R, T> = (selectedEntities?: R[], handlerData?: T) => Promise<string | undefined> | void;
+
+/**
+ * Data required for displaying an action item in a menu.
+ * T is the type of custom data passed to action handler
+ * R is the type of selected entity on which the action will be performed
+ */
+export interface ActionItem<R = unknown, T = unknown> {
```
juanmendes, line 38: Do we really need defaults? I don't see the benefit of defaulting to unknown.

projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * The i18n key for the text of the button
+     */
+    textKey?: string;
+    /**
+     * The css class the button should have.
+     *
+     * @unique among all added buttons
```
juanmendes, line 46: `@unique` is not a jsdoc tag, replace with

Must be unique among all added actions within an action list

PR #114, opened 2020-06-15
Fix selectedOption returning undefined because of mismatching primitives

ps37:
The selectedOption of FormSelect component is returning undefined when there is a mismatch between string and number types of values between option.value and formControl.value even though the values are equal.

For Example: `this.options.find(option => option.value === this.formControl.value);` resulting in...
1000 === "1000" or "1000" === 1000

### DIFF
--- a/projects/components/src/form/form-select/form-select.component.spec.ts
+++ b/projects/components/src/form/form-select/form-select.component.spec.ts
@@ -50,11 +50,11 @@ describe('FormSelectComponent', () => {
         });
         it('is of number type, when the select formControl is set with same number value', () => {
             selectInput.component.formControl.setValue(4);
-            expect(hostComponent.selectInputComponent.selectedOption).toEqual(optionWithValueAsNumber);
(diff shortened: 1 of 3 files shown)


projects/components/src/form/form-select/form-select.component.spec.ts
```diff
@@ -105,3 +112,8 @@ class TestHostComponent {
         });
     }
 }
+
+const optionWithValueAsNumber: SelectOption = {
```
juanmendes, line 116: Please create a function that returns a new object every time instead of passing an object that is likely to be shared between tests

projects/components/src/form/form-select/form-select.component.ts
```diff
@@ -30,6 +30,9 @@ export class FormSelectComponent extends BaseFormControl {
         if (!this.options) {
             return undefined;
         }
-        return this.options.find(option => option.value === this.formControl.value);
+        // option.value is string | number. Also, formControl.value is string when the dropdown value is manually chosen
```
juanmendes, line 33: I think the only comment you need is that the option may be of type number or string

PR #123, opened 2020-06-22
Replace pagination translation call back with a string input

ps37:
This commit has the following changes:
- Make the DG take a string as input for pagination translation instead of a callback

These changes affect VCD UI and they are made in the following MR:
https://gitlab.eng.vmware.com/core-build/vcd_ui/merge_requests/2097

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.ts
+++ b/projects/components/src/datagrid/datagrid.component.ts
@@ -57,6 +57,11 @@ const MAX_HEADER_HEIGHT = 40;
 const ROW_HEIGHT = 37;
 
 /**
+ * Key used for translation of pagination when a translation key is not given as input from the caller
+ */
+export const DEFAULT_PAGINATION_TRANSLATION_KEY = 'vcd.cc.grid.default.pagination';
+
+/**
  * Different types of row selection on the grid
  */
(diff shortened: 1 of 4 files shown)


projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -554,21 +559,9 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit {
     private viewInitted = false;
 
     /**
-     * Gives the correct string to display for the pagination.
-     *
-     * @param firstItem the index of the first item displayed.
-     * @param lastItem the index of the last item displayed.
-     * @param totalItems the total number of items that could be displayed.
+     * Used for translating pagination information displayed in the grid
```
ryan-bradford, line 562: Can you add some information about what data is passed along with the call to `translate`? I.E. firstItem, lastItem, totalItems

ps37: The `getPaginationTranslation` method that actually calls the `translate`, uses `ClrDatagridPagination` given by the HTML and extracts those 3 params from it to pass them to the `translate` method.

As the user did not call the `` method directly with those parameters, I don't think they require to be documented.

PR #127, opened 2020-06-25
Use the same typescript version for @wessberg/ts-evaluator as the route-analyzer

irahov:
If typescripts are different versions then evaluation may not succeed.
Also the typescript library is now a peer dependency since route-analyzer
is not only a standalone tool but also an angular package library,
providing AppRoute interface.
All of this implies that typescript should not be bundled with route-analyzer.

### DIFF
--- a/projects/route-analyzer/package.json
+++ b/projects/route-analyzer/package.json
@@ -1,11 +1,13 @@
 {
   "name": "@vcd/route-analyzer",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "bin": "cli.js",
   "dependencies": {
     "@wessberg/ts-evaluator": "0.0.25",
-    "yargs": "15.3.1",
-    "typescript": "3.5.3"
+    "yargs": "15.3.1"
+  },
+  "peerDependencies": {
+    "typescript": "^3.x"
   },
   "devDependencies": {
     "ts-node": "^8.10.1",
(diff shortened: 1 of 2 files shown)


projects/route-analyzer/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@vcd/route-analyzer",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "bin": "cli.js",
   "dependencies": {
     "@wessberg/ts-evaluator": "0.0.25",
     "yargs": "15.3.1",
-    "typescript": "3.5.3"
+    "typescript": "3.8.3"
```
ps37, line 8: Shouldn’t typescript be listed as a peer dependency as
```
"peerDependencies": {
   "typescript": "^3.x"
},
```
That way, route-analyzer project uses the TS v3.x from base just as it’s transitive dependency @wessberg/ts-evaluator, which also uses TS as peer dependency from the base project.

irahov: I was thinking about that also but on the other side route-analyzer may be used as a stand-alone tool, not depending on the fact that it will be used as part of a base project
it is not like a common component library, where the components have to be part of a project and cannot live by themselves alone - at least these were my considerations

juanmendes: Angular's guidelines are that libraries should not bundle large packages. They should make them peer dependencies, that's why we are required to add dependencies to a white list. See https://github.com/ng-packagr/ng-packagr/blob/master/docs/dependencies.md

Issue #130, opened 2020-07-02
spotlight search styles hides modal-backdrop and modal-headers


ttsvetko:
Please add `:host` in front of `::ng-deep` selector, otherwise it is applied globally.

https://github.com/vmware/vmware-cloud-director-ui-components/blob/master/projects/components/src/spotlight-search/spotlight-search.component.scss#L1

PR #131, opened 2020-07-02
Restrict the css only to the spotlight search component

irahov:
This is to address:
spotlight search styles hides modal-backdrop and modal-headers #130
https://github.com/vmware/vmware-cloud-director-ui-components/issues/130


ryan-bradford:
Juan and I talked and the `:host` selector does not need to be applied to the entire CSS. Remaking this in a new PR where I fix this and bump the version https://github.com/vmware/vmware-cloud-director-ui-components/pull/132

PR #135, opened 2020-07-07
Add Action menu component

ps37:
Add an ActionMenuComponent to display Static(doesn't depend on selected entities) and Contextual actions in either inline bar or a dropdown.
Add a DropdownComponent that can be used for displaying nested dropdowns.

**Why was this change made?**
ActionMenuComponent: Reusable component to display actions in cards, on top of cards, in grid rows, on top of grids, inside entity details screens.
DropdownComponent: Reusable component that was made during the development of the above component. But it can be used for displaying other nested dropdowns and not just actions. Nested dropdowns with a single child will get flattened. For example, like the "Stop/Start" action in the GIF below. They were part of "Power actions" nested dropdown. However, since they are only displayed one at a time, they got flattened.

**Testing Done:**
ActionMenuExampleDemo

**Inside a datagrid:**
ActionMenuInDgExampleDemo

**Jira issue number:** VDUCC-180


projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -118,52 +149,98 @@ export class ActionMenuComponent<R, T> {
...
 
-    private getFlattenedActionList(actions: ActionItem<R, T>[], actionType: ActionType): ActionItem<R, T>[] {
-        let featuredActions: ActionItem<R, T>[] = [];
-        actions.forEach(action => {
+    /**
+     * List containing all the static actions. It has static featured actions in the beginning of the list followed by
+     * non-featured static actions as children of grouped action called 'vcd.cc.action.menu.nonFeatured.actions'
+     */
+    get staticDropdownActions(): ActionItem<R, T>[] | object {
+        return this.staticFeaturedActions.concat([{ textKey: 'Non featured', children: this.staticActions }]);
```
ryan-bradford, line 161: Is 'Non featured' being displayed? if so, it should be translated (either input from the user, or some key that we know about and give a default translation for)

projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -37,7 +37,7 @@ export type ActionHandlerType<R, T> = (selectedEntities?: R[], handlerData?: T)
  */
 export interface ActionItem<R, T> {
     /**
-     * The i18n key for the text of the button
+     * The i18n key or a translated string for contents of a action button
```
ryan-bradford, line 40: What's the use case for passing in already translated text? Isn't that just putting more work on the user? If there is a use case, should we expand this pattern to our other components (namely the data exporter?)

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -629,115 +645,22 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
...
      */
     getMaxFeaturedButtonsOnRow(): number {
         let max = 0;
+        // To reuse the logic of ActionMenuComponent.getContextualFeaturedActions method in the forEach cb below
+        const actionMenuComponent = new ActionMenuComponent();
+        actionMenuComponent.actions = this.actions;
         this.items.forEach(item => {
-            max = Math.max(this.getFeaturedButtons([item]).length, max);
+            actionMenuComponent.selectedEntities = [item];
+            const contextualFeaturedActions = actionMenuComponent.contextualFeaturedActions;
```
ryan-bradford, line 658: Should this be a static method to avoid having to create a fake action menu just to use this getter?

projects/components/src/dropdown/dropdown.component.html
```diff
@@ -0,0 +1,54 @@
+<clr-dropdown class="nested-dropdown">
+    <button
+        [ngClass]="{ 'btn btn-sm btn-link': dropdownTriggerBtnTxt }"
```
ryan-bradford, line 3: What does it mean/how does it act if dropdownTriggerBtnTxt is unset?

projects/components/src/action-menu/action-menu.component.html
```diff
@@ -0,0 +1,84 @@
...
+    [onItemClickedCb]="runActionHandler.bind(this)"
+    [isItemDisabledCb]="isActionDisabled.bind(this)"
+    [dropdownTriggerBtnTxt]="dropdownTriggerBtnText"
+    [dropdownTriggerBtnIcon]="dropdownTriggerBtnIcon"
+    [dropdownItemContents]="actionDisplayConfig.contextual.buttonContents"
+></vcd-dropdown>
+
+<ng-template #actionButton let-action>
+    <button
+        class="btn btn-outline-primary inline-action-button"
```
ryan-bradford, line 70: Should these be displayed as link buttons?

projects/components/src/dropdown/dropdown.component.html
ps37: Those 3 classes will not be applied to the button element when that expression evaluates to false.

projects/components/src/datagrid/datagrid.component.ts
ps37: There are other properties on the `actionMenuComponent` instance that `contextualFeaturedActions` method uses... like`selectedEntities`, `getFlattenedActionList` etc.,
There is no trivial way to just reuse the logic from actionMenuComponent here.

projects/components/src/common/interfaces/action-item.interface.ts
juanmendes: When displaying results, we display a VM name, that is not translatable

projects/components/src/datagrid/datagrid.component.html
```diff
@@ -11,29 +11,21 @@ <h3 *ngIf="this.header" class="vcd-header">{{ header }}</h3>
...
-            <ng-container *ngIf="this.getFeaturedButtons().length === 0">
-                <ng-container *ngTemplateOutlet="dropdown"></ng-container>
-            </ng-container>
-        </div>
+    <clr-dg-action-bar *ngIf="shouldShowActionBarOnTop">
+        <vcd-action-menu
+            [actions]="actions"
+            [selectedEntities]="datagridSelection"
+            [actionDisplayConfig]="actionDisplayConfig"
+            [dropdownTriggerBtnText]="'Actions'"
```
juanmendes: Hard coded English text

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -420,6 +437,28 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
...
+    }
+
+    /**
+     * If the contextual buttons should display in a row.
+     */
+    get shouldDisplayContextualActionsInRow(): boolean {
+        return this.contextualActionPosition && this.contextualActionPosition === ContextualActionPosition.ROW;
+    }
+
+    @Input() actionDisplayConfig: ActionDisplayConfig = DEFAULT_ACTION_DISPLAY_CONFIG;
```
juanmendes, line 460: Inputs and Outputs must be documented

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -310,51 +320,58 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
...
+        return (
+            this.actions.length &&
+            (this.hasStaticActions || (this.hasContextualActions && this.shouldDisplayContextualActionsOnTop))
+        );
+    }
+
+    private get hasStaticActions(): boolean {
+        return (
+            this.actions.filter(
+                action => action.actionType === ActionType.STATIC_FEATURED || action.actionType === ActionType.STATIC
```
juanmendes, line 361:
```ts
return this.actions.some(action => 
    action.actionType === ActionType.STATIC_FEATURED || action.actionType === ActionType.STATIC
)
```

projects/examples/src/components/action-menu/action-menu.examples.module.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { ActionMenuComponent } from '@vcd/ui-components';
+import { Documentation } from '@vmw/ng-live-docs';
+import { ActionMenuExampleComponent } from './action-menu.example.component';
+import { ActionMenuExampleModule } from './action-menu.example.module';
+
+Documentation.registerDocumentationEntry({
+    component: ActionMenuComponent,
+    displayName: 'Action menu',
+    urlSegment: 'ActionMenu',
```
juanmendes, line 16: action-menu

Please remember to be consistent, I've seen some kebabCasing but never TitleCase. Standard for URLs is `whatever-this-is`

projects/examples/src/components/datagrid/datagrid-link.example.component.ts
```diff
@@ -3,136 +3,203 @@
...
+    ContextualActionPosition,
+    DatagridComponent,
     GridColumn,
     GridDataFetchResult,
     GridSelectionType,
     GridState,
-    InactiveButtonDisplayMode,
     TextIcon,
 } from '@vcd/ui-components';
+import { ActionStyling } from '../../../../components/src/common/interfaces';
```
juanmendes, line 19: Examples should not import directly from other projects' source code

projects/examples/src/components/datagrid/datagrid-link.example.component.ts
```diff
@@ -3,136 +3,203 @@
...
+
+type HandlerData = Record[] | Blah;
+
 /**
  * Shows linked buttons on the top of the datagrid.
  * Has examples of both buttons that are global and contextual.
+ *
+ * Enable the delete button to hide it and disable the delete button to show it in disable mode. This is because the
+ * inactive display mode of a button can be set as following:
+ * Hidden - availability is false and disabled is false
```
juanmendes, line 39: This should be a list

```ts
/** 
 * ...
 * Enable the delete button to hide it and disable the delete button to show it in disable mode. This is because the
 * inactive display mode of a button can be set as following:
 *  - Hidden - availability is false and disabled is false
 *  - Shown as Disabled - availability is false and disabled is true
 * Please refer to the delete action object configuration in the actions object list configuration.
```

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -174,9 +251,55 @@ export class ActionMenuComponent<R, T> {
...
+    actionsTrackBy: TrackByFunction<ActionItem<R, T>> = (index: number, item: ActionItem<R, T>): string => {
+        return item.textKey;
+    };
+
+    // Following methods are used to avoid logic inside the HTML template
+    /**
+     * To show or hide the container elements containing inline and also dropdown actions
+     */
+    shouldDisplayActions(style: ActionStyling): boolean {
+        return (
```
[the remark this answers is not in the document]
ps37: Those parens are being added by Prettier hook

projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -84,6 +84,11 @@ export interface ActionItem<R, T> {
      * List of actions that will be grouped under this action
      */
     children?: ActionItem<R, T>[];
+
+    /**
+     * To mark if the {@link #ActionItem.textKey} does not require translation
+     */
+    isNotTranslatable?: boolean;
```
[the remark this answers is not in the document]
ps37: My intention is that the text keys of the ActionItems are translatable by default, as most of the action button texts have to be translated currently in our VCD UI. If someone wants to pass in a an action item that already has a translated text, only they have to provide this property.

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -20,23 +20,33 @@ import {
...
     GridColumn,
     GridColumnHideable,
-    InactiveButtonDisplayMode,
 } from './interfaces/datagrid-column.interface';
-import { ContextualButton } from './interfaces/datagrid-column.interface';
+
+/**
+ * An enum that describes where the contextual buttons should display.
+ */
+export enum ContextualActionPosition {
```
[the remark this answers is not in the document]
ps37: Card view doesn't have to worry about this. Since action menu is part of our common datagrid component, the caller has to tell the datagrid where to display the action menu component.

In the case of card, the actions are always inside the card. If they have to be displayed on the top, that is the responsibility of card container component.

projects/examples/src/components/action-menu/action-menu.example.component.ts
```diff
@@ -0,0 +1,198 @@
...
+    `,
+})
+export class ActionMenuExampleComponent<R extends Record, T extends HandlerData> {
+    actions: ActionItem<R, T>[] = [
+        {
+            textKey: 'Static Featured 1',
+            handler: () => console.log('Static Featured 1'),
+            availability: () => true,
+            actionType: ActionType.STATIC_FEATURED,
+            isNotTranslatable: true,
```
juanmendes: Let's avoid the double negative. You can either:

* Name it `isTranslatable` and default it to true if its value is undefined wherever it's used. It's only used in a single place but would get ugly if we have to read it in multiple places
* Make it a class instead of an interface and default it from the constructor
* Rename it to `isRawString`

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -53,26 +42,18 @@ export class ActionMenuComponent<R, T> {
...
-     * Display configuration of static and contextual actions
+     * Display configuration of both static and contextual actions
      * If null or undefined is passed, default config {@link _actionDisplayConfig} is used
      */
     @Input() set actionDisplayConfig(config: ActionDisplayConfig) {
-        config = !config ? this._actionDisplayConfig : config;
-        Object.keys(config).forEach(key => {
-            this._actionDisplayConfig[key] = config[key] || this._actionDisplayConfig[key];
-        });
+        Object.keys(config || {}).forEach(
```
juanmendes, line 54: Are you forgetting to reset `this._actionDisplayConfig = {}`?

projects/components/src/dropdown/dropdown.component.html
```diff
@@ -0,0 +1,61 @@
...
+                <ng-container *ngIf="!action.children">
+                    <button
+                        class="btn btn-link"
+                        [ngClass]="{ 'btn-icon': shouldShowIcon }"
+                        clrDropdownItem
+                        (click)="onItemClickedCb(action)"
+                        [disabled]="isItemDisabledCb(action)"
+                    >
+                        <ng-container *ngIf="shouldShowText">{{
+                            action.isTranslatable == null || action.isTranslatable === true
```
juanmendes, line 45:
```ts
if (action.isTranslatable === false) {

} else {

}

File: projects/components/src/action-menu/action-menu.component.html [hunk 18]
```diff
@@ -0,0 +1,80 @@
+<div class="inline-actions-container" *ngIf="shouldDisplayActions(actionStyling.INLINE)">
```
irahov, line 1: Is there other cases where `shouldDisplayActions` is called with other(different) parameters? If not - then why not using a property that is calculated just once

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -310,51 +322,56 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
...
+     * Actions to be displayed on a grid or in a row
+     */
+    private _actions: ActionItem<R, unknown>[] = [];
+
+    /**
+     * List of actions given by the caller
+     */
+    @Input() set actions(actions: ActionItem<R, unknown>[]) {
+        this._actions = actions.map(action => {
+            const actionHandler: ActionHandlerType<R, unknown> = action.handler;
```
irahov, line 334: It is not good to modify the provided element, better do :
action = {...action}

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -310,51 +322,56 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
...
+        });
+    }
+    get actions(): ActionItem<R, unknown>[] {
+        return this._actions;
+    }
+
+    /**
+     * To display static and contextual actions in a action bar on top of a grid
+     */
+    get shouldShowActionBarOnTop(): boolean {
```
irahov, line 351: @juanmendes 
maybe this is one of the reasons thy he battery of my laptop is drained out with our app. On any angular 'digest' cycle this function will be called over and over again. Internally it will call hasStaticActions which iterates through an array and so on and so forth

Issue #134, opened 2020-07-06
Unable to set `unlimited` value as default one for `vcd-number-with-unit-form-input`form component


ttsvetko:
I need to set `unlimited` option by default for that component. I set `unlimitedValue` to `-1` and then i set the value of the formControl to `-1` but `unlimited` is not checked.

https://stackblitz.com/edit/angular-3ekvtd?file=src%2Fapp%2Fnumber-with-unit-form-input.example.component.html

PR #137, opened 2020-07-08
Fix bug where unable to set 'unlimited' value as default one (vmware#…

yumengwu:
…134) Fix the bug to enable unlimited as default.

Fixes: #134

### DIFF
--- a/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.spec.ts
+++ b/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.spec.ts
@@ -27,12 +27,23 @@ const [MIN, MAX] = [1000, 10000];
             [min]="${MIN}"
             [max]="${MAX}"
             [description]="'sizingPolicies.form.cpuLimit.description' | translate"
+            class="initially-null"
+        >
+        </vcd-number-with-unit-form-input>
+        <vcd-number-with-unit-form-input
+            [formControl]="cpuLimitInitiallyUnlimited"
+            [label]="'cpu.limit' | translate"
+            [unitOptions]="hertzOptions"
+            [inputValueUnit]="formControlValueUnit"
+            [description]="'sizingPolicies.form.cpuLimit.description' | translate"
+            class="initially-unlimited"
         >
         </vcd-number-with-unit-form-input>
     `,
 })
 export class TestHostComponent {
     cpuLimit = new FormControl(null);
+    cpuLimitInitiallyUnlimited = new FormControl(UNLIMITED);
 
     public hertzOptions: Unit[] = [Hertz.Mhz, Hertz.Ghz];
     public formControlValueUnit: Unit = Hertz.Mhz;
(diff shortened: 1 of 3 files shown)


projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.spec.ts
```diff
@@ -143,6 +154,22 @@ describe('VcdNumberWithUnitFormInputComponent', () => {
...
+        it('disables text input when initialized to unlimited', () => {
+            const numberWithUnitInputInitializedUnlimited = finder.find({
+                woConstructor: NumberWithUnitFormInputWidgetObject,
+                className: 'second',
+            });
+            expect(numberWithUnitInputInitializedUnlimited.valueFormControl.enabled).toEqual(false);
+            expect(numberWithUnitInputInitializedUnlimited.unitFormControl.enabled).toEqual(false);
+        });
+
+        it('disables text input when set unlimited programmatically', () => {
```
juanmendes, line 167: Typo: "disables text input when set to unlimited programmatically"

projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.ts
```diff
@@ -185,21 +185,21 @@ export class NumberWithUnitFormInputComponent extends BaseFormControl implements
...
             this.formGroup = this.fb.group({
-                limited: [this.bestValue, validatorUnlimited],
+                limited: [this.bestValue === -1 ? 0 : this.bestValue, validatorUnlimited],
                 comboUnitOptions: this.bestUnit.getMultiplier(),
-                unlimited: false,
+                unlimited: this.bestValue === -1 ? true : false,
             });
             this.tracker.subscribe(this.formGroup.get('unlimited').valueChanges, value => {
                 const input = this.formGroup.get('limited');
                 const comboUnitOptions = this.formGroup.get('comboUnitOptions');
```
juanmendes, line 194: `comboUnitOptions ` is not being used

projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.ts
```diff
@@ -185,21 +185,21 @@ export class NumberWithUnitFormInputComponent extends BaseFormControl implements
...
-                    comboUnitOptions.enable();
+                // If checkbox is limited and value is -1, or if checkbox is unlimited and input value is empty,
+                // set input value to 0.
+                if ((value === false && input.value === -1) || (value === true && input.value)) {
+                    input.setValue(0);
                 }
-                this.onChange(this.value());
+                const inputValue = this.value();
+                this.formControl.setValue(inputValue ? inputValue : 0, { emitModelToViewChange: true });
+                // this.onChange(this.value());
```
juanmendes, line 202: Don't leave uncommented code in. Please remove it.

Right now, it's a bit unclear as to why we're doing something different here. We could add comments to explain but we could also rework this.

Bring back `this.onChange(this.value)`

After calling `this.onChange`, we call a new private method that enabled/disables fields based on whether it's unlimited. That new function will be  called from `writeValue()` and from here.

Example:

```ts

// Makes sure that the input fields are disabled when this widget is set to UNLIMITED
private updateUiDisabledState() {
        if (value === UNLIMITED) {
            input.disable();
            comboUnitOptions.disable();
        } else {
            input.enable();
            comboUnitOptions.enable();
            input.setValue(this.bestValue);
            comboUnitOptions.setValue(this.bestUnit.getMultiplier());
        }
}
```


```

PR #141, opened 2020-07-20
Add urlSegment to each example.

yumengwu:
Add urlSegment variable to each example's exampleEntry to enable live-docs' sub-route.

**Why was this change made?**
There is one more required property in live-docs: urlSegment, which is used to enable sub-route for each example.

**Testing Done:**
ui-component-test-done


juanmendes:
It looks like it's failing because you forgot to publish live-docs and reference the new version in package.json

juanmendes:
From the error, it looks like you didn't update package-lock. To update a dependency, you should use `npm install @some/lib-name 1.0.0.dev-002` --save-exact instead of modifying package.json manually. npm install will update package-lock.json


image

PR #143, opened 2020-07-21
Action search provider for registering actions with quick search

ps37:
**Why was this change made?**
To make actions given to the action menu component be available for searching on quick search component

**What was the change?**
Make the ActionMenuComponent as a SpotlightSearchProvider so that it can search through its nested actions

**Testing Done:**
The component was used inside EntityDetailsContainer of vcd_ui and verified that the actions available on the entity inside the details screen can be searched and performed through quick search.
ActionSearch

**Jira task Number:**
https://jira.eng.vmware.com/browse/VDUCC-181


projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -96,6 +111,38 @@ export class ActionMenuComponent<R, T> {
      */
     @Input() nestedDropdownPosition = 'right-top';
 
+    private _isActionSearchProvider = false;
+    /**
+     * This component is NOT registered by default with with the {@link SpotlightSearchService} to register its actions to show up on the
+     * {@link SpotlightSearchComponent}. However, in the VCD app, some components like the entity details container want to register their
```
irahov, line 117: We should not reference vcd specific stuff like the entity details container

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -96,6 +111,38 @@ export class ActionMenuComponent<R, T> {
...
+    }
+    get isActionSearchProvider(): boolean {
+        return this._isActionSearchProvider;
+    }
+
+    private _actionProviderName: string;
+    /**
+     * Is displayed on the quick search modal as Actions: <actionProviderName>
+     */
+    @Input() set actionProviderName(val: string) {
```
irahov, line 136: In general - if 2 properties go together better group them.
In this specific case it is a bit tedious to know all those interdependencies that in order to provide name, you have to set also `isActionSearchProvider`.

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private findSearchCriteriaInNestedActions(
+        criteria: string,
+        actions: ActionItem<R, T>[] = this.actions,
+        matchingResults: ActionItem<R, T>[] = []
+    ): ActionItem<R, T>[] {
+        const nonDisabledActions = actions.filter(action => !this.isActionDisabled(action));
+        nonDisabledActions.forEach(action => {
+            if (action.children && action.children.length) {
```
irahov, line 87: single check in TS: action?.children?.length

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,99 @@
...
+                this.findSearchCriteriaInNestedActions(criteria, action.children, matchingResults);
+            } else if (action.textKey.toLowerCase().includes(criteria)) {
+                matchingResults.push(action);
+            }
+        });
+        return matchingResults;
+    }
+
+    private isActionDisabled(action: ActionItem<R, T>): boolean {
+        return typeof action.disabled === 'function' ? action.disabled(this.selectedEntities) : action.disabled;
```
irahov, line 97: 1) Maybe we can add such a utility isFunction()
2) Is it possible action.disabled to change after initialization

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,99 @@
...
+        this.spotlightSearchService.unregisterProvider(this.spotlightSearchProviderRegistrationId);
+    }
+
+    private findSearchCriteriaInNestedActions(
+        criteria: string,
+        actions: ActionItem<R, T>[] = this.actions,
+        matchingResults: ActionItem<R, T>[] = []
+    ): ActionItem<R, T>[] {
+        const nonDisabledActions = actions.filter(action => !this.isActionDisabled(action));
+        nonDisabledActions.forEach(action => {
```
irahov, line 86: What is the case when action is 'actionable' AND can have children in the same time?

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,99 @@
...
+            displayText: action.isTranslatable === false ? action.textKey : this.ts.translate(action.textKey),
+            handler: action.handler,
+        }));
+    }
+
+    /**
+     * Registers the actions with {@link SpotlightSearchService}. Called by {@link ActionMenuComponent}
+     */
+    register(): void {
+        if (this.spotlightSearchProviderRegistrationId) {
```
irahov, line 60: As it is of now it seems that we cannot instantiate 2 ActionSearchProvider-s. If for some reason there are 2 components in the UI the second one that injects the ActionSearchProvider will unregister the previous one

projects/examples/src/components/action-menu/action-menu.example.component.ts
```diff
@@ -192,4 +206,14 @@ export class ActionMenuExampleComponent<R extends Record, T extends HandlerData>
     toggleDropdownDisable(): void {
         this.isDropdownDisabled = !this.isDropdownDisabled;
     }
+
+    ngOnInit(): void {
+        mousetrap.bind(this.kbdShortcut, () => {
+            this.spotlightOpen = true;
+            // Since this happens outside angular zone we need to notify angular manually
+            // otherwise the user should wait for any operation that will trigger angular change detection (angular zone event)
+            this.changeDetectorRef.detectChanges();
```
irahov, line 215: If you take a look in the core vcd-ui code there is a way to avoid this change detection. Basically by createin a new instance of the mousetrap in the constructor

projects/components/src/action-menu/action-search.provider.ts
ps37: This provider is provided for each action menu component instance. Please see https://github.com/vmware/vmware-cloud-director-ui-components/pull/143/files#diff-1ae5d00fd1a984540f5a721b704076b7R32

projects/components/src/action-menu/action-search.provider.ts
ps37: 1. Done
2. It can

projects/components/src/action-menu/action-search.provider.ts
ps37: An actions item that has children is not actionable.

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,120 @@
...
+        return this._actions;
+    }
+
+    /**
+     * Used in {@link isActionDisabled} to calculate disabled state of actions
+     */
+    private _selectedEntities: R[] = [];
+    set selectedEntities(entities: R[]) {
+        this._selectedEntities = entities;
+        this.updateFlatListOfAvailableActions();
```
irahov, line 34: Isn't it better to this lazily when doing the first search? Otherwise on any selection change, we will do extra computation even if not necessary

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,120 @@
+/*!
```
irahov, line 1: Better not to place this class under action-menu, since it is not related to it, but rather to the ActionItem only

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            const spy = spyOn((this.actionSearchProvider as any).spotlightSearchService, 'registerProvider');
+            this.actionSearchProvider.register();
+            expect(spy).toHaveBeenCalledWith(
+                this.actionSearchProvider,
+                '{"key":"vcd.cc.action.provider.section.title","params":[{"actionProviderName":""}]}'
+            );
+        });
+
+        it(
+            'calls the registerProvider method of spotlight search service if registration id exists' + ' already',
```
irahov, line 81: 1) you mean unregister?
2) the test should not reveal private details like if registration id exists or not. What is important that if the name changes this will be reflected in the spotlight search service - so you can test that changing the name results in unregister and register

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,120 @@
...
+     * Searches through nested actions and finds all the actions that match with entered search text on the
+     * {@link SpotlightSearchComponent}
+     */
+    search(criteria: string): SpotlightSearchResult[] {
+        if (!criteria) {
+            return [];
+        }
+
+        return this.findActions(criteria.toLowerCase()).map(action => ({
+            displayText: action.isTranslatable === false ? action.textKey : this.ts.translate(action.textKey),
```
irahov, line 69: the display text is not going to be changed, so no need to do this in every search

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            function(this: HasActionSearchProvider): void {
+                const spy = spyOn((this.actionSearchProvider as any).spotlightSearchService, 'unregisterProvider');
```
irahov, line 83: Do not get the spotlightSearchService from the private property of the provider. In general getting access to private properties or methods should be a very rare case

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+                (this.actionSearchProvider as any).spotlightSearchProviderRegistrationId = 'some_id';
+                this.actionSearchProvider.register();
+                expect(spy).toHaveBeenCalled();
+            }
+        );
+    });
+
+    describe('unregister', () => {
+        it('calls the registerProvider method of spotlight search service', function(this: HasActionSearchProvider): void {
```
irahov, line 92: unregisterProvider

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            const spy = spyOn((this.actionSearchProvider as any).spotlightSearchService, 'unregisterProvider');
+            (this.actionSearchProvider as any).spotlightSearchProviderRegistrationId = 'some_id';
```
irahov, line 94: Just register it first, do not play with private properties

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            const spy = spyOn(this.actionSearchProvider as any, 'updateFlatListOfAvailableActions').and.callFake(
+                () => null
+            );
+            this.actionSearchProvider.selectedEntities = [];
+            expect(spy).toHaveBeenCalled();
+        });
+    });
+
+    describe('actionProviderName', () => {
+        it('calls the register method when set', function(this: HasActionSearchProvider): void {
```
irahov, line 63: It is not important that setting the provider name will call register or not. It is important that this name willl be reflected in the spotlight search service

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+describe('ActionSearchProvider', () => {
+    beforeEach(function(this: HasActionSearchProvider): void {
+        this.actionSearchProvider = new ActionSearchProvider(
+            new SpotlightSearchService(),
+            new MockTranslationService()
+        );
+    });
+    describe('actions', () => {
+        it('updates the flattened list of available actions when set', function(this: HasActionSearchProvider): void {
```
irahov, line 43: I don't think that the test should explain the inner structure. They should explain what is the behavior of the item in terms of its public behavior

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            expect((this.actionSearchProvider as any).flatListOfAvailableActions).toEqual([]);
+            this.actionSearchProvider.actions = MOCK_DATA;
+            expect((this.actionSearchProvider as any).flatListOfAvailableActions).toEqual(
+                (this.actionSearchProvider as any).getFlatListOfAvailableActions(MOCK_DATA)
+            );
+        });
+    });
+
+    describe('selectedEntities', () => {
```
irahov, line 52: selectedEntities are important in checking action availability and  the action handler, not the private behavior of flattening the array

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+import { MockTranslationService } from '@vcd/i18n';
+import { ActionItem } from '../common/interfaces/index';
+import { SpotlightSearchService } from '../spotlight-search/index';
+import { ActionSearchProvider } from './action-search.provider';
+
+interface HasActionSearchProvider {
+    actionSearchProvider: ActionSearchProvider<any, any>;
+}
+
+const MOCK_DATA: ActionItem<any, any>[] = [
```
irahov, line 15: Better have this as a function so that it gets not modified accidentally (in the future when new test may be added)

projects/components/src/action-menu/action-search.provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { MockTranslationService } from '@vcd/i18n';
+import { ActionItem } from '../common/interfaces/index';
+import { SpotlightSearchService } from '../spotlight-search/index';
+import { ActionSearchProvider } from './action-search.provider';
+
+interface HasActionSearchProvider {
```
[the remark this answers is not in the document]
ps37: I followed the convention at other places like `HasFinder`, `HasFinderAndGrid` etc.,

projects/components/src/action-menu/action-search.provider.spec.ts
ps37: But that is not the job of the `ActionSearchprovider` to make sure about that right. Once it is passed to the SpotlightSearch service it is the job of that to make sure about this

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Registers the actions with {@link SpotlightSearchService}. Called by {@link ActionMenuComponent}
+     */
+    register(): void {
+        if (this.spotlightSearchProviderRegistrationId) {
+            this.spotlightSearchService.unregisterProvider(this.spotlightSearchProviderRegistrationId);
+        }
+        this.spotlightSearchProviderRegistrationId = this.spotlightSearchService.registerProvider(
+            this,
+            this.ts.translate(ACTION_PROVIDER_SECTION_TITLE_KEY, [
```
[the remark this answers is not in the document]
ps37: When we were doing POC testing, Stephen Evanchik mentioned that the heading for actions section in English language should follow this convention `Actions: {actionProviderName}`

projects/components/src/action-menu/action-search.provider.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Un registers the actions with {@link SpotlightSearchService}. Called by {@link ActionMenuComponent}
+     */
+    unregister(): void {
+        this.spotlightSearchService.unregisterProvider(this.spotlightSearchProviderRegistrationId);
+    }
+
+    private findActions(searchCriteria: string): ActionItem<R, T>[] {
+        return this.flatListOfAvailableActions.filter(action => action.textKey.toLowerCase().includes(searchCriteria));
```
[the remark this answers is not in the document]
ps37: Good catch!

projects/components/src/action-menu/action-search.provider.ts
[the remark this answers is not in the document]
ps37: But you are right in that maybe another app than VCD might want it in a different format

projects/examples/src/components/action-menu/action-menu.example.component.ts
```diff
@@ -44,7 +55,8 @@ interface HandlerData {
             <button (click)="toggleDropdownDisable()" class="btn btn-primary">
                 {{ isDropdownDisabled ? 'Enable dropdown' : 'Disable dropdown' }}
             </button>
-            <br />
+            <br /><br />
+            <p>Press 'mod+.' on the keyboard to open quick search and search for contextual actions</p>
```
irahov, line 59: I'm about to change this at that other place. Sorry for providing bad source to copy from :( 
'mod' is internal. Better say Press 'command+.' on Mac OS /'ctrl+.' on PC/ ...

projects/examples/src/components/action-menu/action-menu.example.component.ts
```diff
@@ -54,9 +66,38 @@ interface HandlerData {
...
+            [(open)]="spotlightOpen"
+            [placeholder]="'Search contextual actions'"
+        ></vcd-spotlight-search>
     `,
 })
-export class ActionMenuExampleComponent<R extends Record, T extends HandlerData> {
+export class ActionMenuExampleComponent<R extends Record, T extends HandlerData> implements OnInit, OnDestroy {
+
+    constructor(
+        private changeDetectorRef: ChangeDetectorRef,
```
irahov, line 79: you don't need: changeDetectorRef

projects/components/src/action-search-provider/action-search.provider.ts
```diff
@@ -0,0 +1,87 @@
...
+            this.flatListOfAvailableActions = this.getFlatListOfAvailableActions(this.actions);
+        }
+
+        return this.findActions(criteria.toLowerCase()).map(action => ({
+            displayText: action.textKey,
+            handler: () => action.handler(this.selectedEntities, action.handlerData),
+        }));
+    }
+
+    private findActions(searchCriteria: string): ActionItem<R, T>[] {
```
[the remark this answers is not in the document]
ps37: Kept the method as I feel it makes the class a little more readable. However, added more logic to it.

projects/components/src/action-search-provider/action-search.provider.ts
```diff
@@ -0,0 +1,87 @@
...
+    private getFlatListOfAvailableActions(actions: ActionItem<R, T>[]): ActionItem<R, T>[] {
+        return actions
+            .reduce((flatActionList, currentAction) => {
+                if (currentAction?.children?.length) {
+                    flatActionList = flatActionList.concat(this.getFlatListOfAvailableActions(currentAction.children));
+                } else if (this.isActionAvailable(currentAction)) {
+                    const currentActionCopy = { ...currentAction };
+                    currentActionCopy.textKey = currentAction.isTranslatable === false ? currentAction.textKey :
+                        this.ts.translate(currentAction.textKey);
+                    flatActionList.push(currentActionCopy);
```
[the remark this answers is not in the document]
ps37: I agree that your suggestion is more expressive about what the code is doing. I took it!

projects/components/src/utils/common-util.ts
```diff
@@ -15,4 +15,11 @@ export class CommonUtil {
         }
         return Number(Math.round(Number(value + 'e' + digits)) + 'e-' + digits);
     }
+
+    /**
+     * Verifies if the passed value is of type Function
+     */
+    static isFunction(value: any): value is (...args: unknown[]) => unknown {
```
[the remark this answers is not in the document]
ps37: I learned that from Juan!

PR #147, opened 2020-07-28
Spotlight search

irahov:
Please take a look at the different commits for more details


projects/examples/src/components/spotlight-search/spotlight-search-example.component.ts
```diff
@@ -35,6 +35,8 @@ export class SpotlightSearchExampleComponent implements OnInit, OnDestroy {
             ['placeholder']: [''],
         });
 
+        this.actionsSearchProvider.register();
```
ryan-bradford, line 38: Doesn't really matter because it's an example, but should this be moved to ngOnInit?

PR #155, opened 2020-07-30
Created new WidgetObject pattern for Cypress and Angular tests

ryan-bradford:
# Description

Creates a new WidgetObject pattern so that widget objects can be used in both Cypress and Angular tests. All new WidgetObjects should now extend the BaseWidgetObject which will give all the necessary methods to test features.

# Design

1. WidgetDriver interface
- this is the interface for an object that knows how to interact with the dom. Currently we have three CypressWidgetDriver (cypress), AngularWidgetDriver (angular), BaseWidgetDriver (delegation to another).
2. BaseWidgetDriver
- All new WidgetObjects should extend this to gain the ability to interact with the dom for any arbitrary testing framework.
3. WidgetFinders(s)
- Given a FindParams, locates the specific WidgetDriver and instantiates the WidgetObject. Two Widget Finders, one for Cypress, one for Angular

# Testing
1. Added the Angular Widget Driver spec
2. Implemented the rights bundle widget using the BaseWidgetDriver in VCD-UI

# Notes
1. The `find` method is not useful for testing if a non-existent element has properties. For example, we can't use find to test things about the Spinner because when the spinner is off, it's removed from the dom. Will discuss in standup tomorrow.

### DIFF
--- a/projects/components/src/utils/test/angular-widget-driver.ts
+++ b/projects/components/src/utils/test/angular-widget-driver.ts
@@ -1,0 +1,119 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { DebugElement } from '@angular/core';
+import { ComponentFixture } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { AngularWidgetFinder } from './angular-widget-finder';
+import { WidgetDriver } from './base-widget-object';
+import { FindableWidget, FindParams } from './widget-object';
+
+/**
+ * A WidgetDriver that knows how to interface with the Angular framework.
+ */
+export class AngularWidgetDriver<T> implements WidgetDriver {
+    constructor(
+        protected fixture: ComponentFixture<any>,
+        protected root: DebugElement = fixture.debugElement,
+        protected component: T = fixture.componentInstance
+    ) {}
+
+    find<W extends WidgetDriver>(params: FindParams<FindableWidget<W>> | FindableWidget<W>): W {
+        return new AngularWidgetFinder(this.fixture).find<W, FindableWidget<W>>(params, this.root);
+    }
+
+    async click(cssSelector?: string): Promise<void> {
+        this.getDebugElement(cssSelector).nativeElement.click();
+        this.fixture.detectChanges();
+    }
+
+    async classes(cssSelector?: string): Promise<string[]> {
+        return Object.keys(this.getDebugElement(cssSelector).nativeElement.classes);
+    }
+
+    async check(cssSelector?: string): Promise<void> {
+        const checkboxElement = this.getDebugElement(cssSelector).nativeElement;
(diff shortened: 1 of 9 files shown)


projects/components/src/utils/test/angular-widget-driver.ts
```diff
@@ -0,0 +1,122 @@
...
+        this.getDebugElement(cssSelector).nativeElement.click();
+        this.fixture.detectChanges();
+        return Promise.resolve();
+    }
+
+    classes(cssSelector?: string): Promise<string[]> {
+        return Promise.resolve(Object.keys(this.getDebugElement(cssSelector).nativeElement.classes));
+    }
+
+    check(cssSelector?: string): Promise<void> {
```
juanmendes, line 37: Mark your functions as `async` and you don't need to return  a Promise. That should clean the code a lot.

projects/components/src/utils/test/cypress-widget-driver.ts
```diff
@@ -0,0 +1,145 @@
...
+                .check()
+                .then(() => resolve())
+        );
+    }
+
+    getChecked(cssSelector?: string): Promise<boolean> {
+        return this.assert('be.checked', cssSelector);
+    }
+
+    select(option: string, cssSelector?: string): Promise<void> {
```
juanmendes, line 67: This should be made `async` as I mentioned. But why didn't you just `return this.getElement(cssSelector).select(option)` ? Usually, creating Promise is an anti-pattern. See https://stackoverflow.com/questions/23803743/what-is-the-explicit-promise-construction-antipattern-and-how-do-i-avoid-it

juanmendes:
About your Notes section. Do you mean cypress doesn't allow us to test that something is not present in the dom?

projects/components/src/utils/test/cypress-widget-driver.ts
```diff
@@ -7,6 +7,27 @@ import { WidgetDriver } from './base-widget-object';
 import { CypressWidgetFinder } from './cypress-widget-finder';
 import { FindableWidget, FindParams } from './widget-object';
 
+declare const Cypress;
+
+export function promisify<T>(chain): Promise<T> {
```
juanmendes, line 12: Please add comments, exported/public members should all have some docs

projects/components/src/utils/test/angular-widget-driver.ts
```diff
@@ -24,90 +24,81 @@ export class AngularWidgetDriver<T> implements WidgetDriver {
...
     }
 
-    clearInputValue(cssSelector?: string): Promise<void> {
+    async clearInputValue(cssSelector?: string): Promise<void> {
         return this.setInputValue('', cssSelector);
     }
 
-    getText(cssSelector?: string): Promise<string> {
-        return this.getTexts(undefined, cssSelector).then(texts => texts[0]);
+    async getText(cssSelector?: string): Promise<string> {
```
juanmendes, line 72: This should be implemented using `getDebugElement()` for performance reasons (so it stops searching after it finds one)

PR #160, opened 2020-08-05
Add class name to action items VcdDropdown

ps37:
# Why?
The end 2 end tests in vcd ui are failing because of missing class names on action items inside action dropdowns. So, added those class names

# Testing:
Added class names to action item objects in ActionMenuExample component and made sure that those classes get added to the class list of action button html element

### DIFF
--- a/projects/components/src/action-menu/action-menu.component.ts
+++ b/projects/components/src/action-menu/action-menu.component.ts
@@ -43,7 +43,8 @@ export class ActionMenuComponent<R, T> {
         const hasNestedActions = copyOfActions.some(action => action.children && action.children.length);
         const shouldDisplayContextualActionsDropdown =
             hasNestedActions ||
-            copyOfActions.some(action => action.actionType && action.actionType === ActionType.CONTEXTUAL_FEATURED);
+            this.getFlattenedActionList(copyOfActions, ActionType.CONTEXTUAL_FEATURED)
+                .some(action => action.actionType && action.actionType === ActionType.CONTEXTUAL_FEATURED);
 
         this._actions = copyOfActions.map(action => {
             if (!action.actionType) {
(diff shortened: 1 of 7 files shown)


projects/components/src/dropdown/dropdown.component.html
```diff
@@ -36,6 +36,7 @@
                 <ng-container *ngIf="!action.children">
                     <button
                         class="btn btn-link"
+                        [className]="action.class"
```
juanmendes, line 39: We're already using `ngClass`, why use something else?

projects/components/src/dropdown/dropdown.component.ts
```diff
@@ -27,6 +27,10 @@ interface DropdownItem<T extends DropdownItem<T>> {
      * To mark if the {@link #ActionItem.textKey} has to be translated or not
      */
     isTranslatable?: boolean;
+    /**
+     * Css class of the dropdown item. Must be unique among all dropdown items within the dropdown items list
+     */
+    class: string;
```
juanmendes, line 33: please name this className, class is a reserved word, it's allowed here but people tend to avoid it to minimize confusion

projects/components/src/dropdown/dropdown.component.html
ps37: ngClass lets me add a string value conditionally. But here, I want to ass a property value as a class name. that is why I did this.

projects/components/src/dropdown/dropdown.component.ts
ps37: I named it to class by following the existing ActionItem interface in vcd ui to avoid a big refactor. I use the class property coming form the action item object in this component. If I change this to className, I have to either change the action item object as well(a big refactor in our repo and also vcd ui repo) or do some work inside this component top convert class -> className.

juanmendes: OK

projects/components/src/dropdown/dropdown.component.html
juanmendes: It seems we should replace class with 

```
[className]="[action.class, 'btn', 'btn-link']"
```

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -65,14 +66,16 @@ export class ActionMenuComponent<R, T> {
...
+    private _actionDisplayConfig: ActionDisplayConfig = getCopyOfActionDisplayConfig(DEFAULT_ACTION_DISPLAY_CONFIG);
     /**
      * Display configuration of both static and contextual actions
      * If null or undefined is passed, default config {@link _actionDisplayConfig} is used
      */
     @Input() set actionDisplayConfig(config: ActionDisplayConfig) {
-        Object.keys(config || {}).forEach(
-            key => (this._actionDisplayConfig[key] = config[key] ? config[key] : DEFAULT_ACTION_DISPLAY_CONFIG[key])
+        const copyOfConfig = getCopyOfActionDisplayConfig(config || {});
+        const copyOfDefaultConfig = getCopyOfActionDisplayConfig(DEFAULT_ACTION_DISPLAY_CONFIG);
```
juanmendes, line 76: Instead of duplicating code to get a copy, create a function to get a new instance every time

```
// Hard code objects returned will always be new instances
getDefaultActionDisplayConfig() {

}
```

PR #162, opened 2020-08-06
Make changes to not mutate action display config objects

ps37:
# Why?
- To avoid mutating _actionDisplayConfig being used internally by the ActionMenuComponent to control the view
- To avoid mutating the config object being passed as input from the calling component using ActionMenuComponent

# Testing done:
Added unit tests to test the above 2 scenarios

### DIFF
--- a/projects/components/src/action-menu/action-menu.component.spec.ts
+++ b/projects/components/src/action-menu/action-menu.component.spec.ts
@@ -6,9 +6,9 @@
 import { Component } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { MockTranslationService, TranslationService } from '@vcd/i18n';
-import { ActionStyling, ActionType, TextIcon } from '../common/interfaces/index';
+import { ActionDisplayConfig, ActionStyling, ActionType, TextIcon } from '../common/interfaces/index';
 import { WidgetFinder, WidgetObject } from '../utils/test/index';
-import { ActionMenuComponent } from './action-menu.component';
+import { ActionMenuComponent, getDefaultActionDisplayConfig } from './action-menu.component';
 import { VcdActionMenuModule } from './action-menu.module';
 
 interface HasFinderAndActionMenu {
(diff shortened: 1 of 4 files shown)


projects/components/src/action-menu/action-menu.component.spec.ts
```diff
@@ -20,7 +20,7 @@ export class ActionMenuWidgetObject<R, T> extends WidgetObject<ActionMenuCompone
     static tagName = 'vcd-action-menu';
 }
 
-describe('ActionMenuComponent', () => {
+fdescribe('ActionMenuComponent', () => {
```
ryan-bradford, line 23: Remove fdescribes

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -10,14 +10,16 @@ import { CommonUtil } from '../utils';
...
  */
-export const DEFAULT_ACTION_DISPLAY_CONFIG: ActionDisplayConfig = {
-    contextual: {
-        featuredCount: 0,
-        styling: ActionStyling.INLINE,
-        buttonContents: TextIcon.TEXT,
-    },
-    staticActionStyling: ActionStyling.INLINE,
-};
+export function getDefaultActionDisplayConfig(): ActionDisplayConfig {
```
juanmendes, line 13: Calling code will be simpler if you take in a cfg object and apply the defaults on top of it instead of letting callers do it.

```ts
/**
 * Value used for the display configuration of action buttons if no input is provided by the caller
 */
export function getDefaultActionDisplayConfig(cfg: ActionDisplayConfig = {}): ActionDisplayConfig {
    const defaults =  {
        contextual: {
            featuredCount: 0,
            styling: ActionStyling.INLINE,
            buttonContents: TextIcon.TEXT,
        },
        staticActionStyling: ActionStyling.INLINE,
    };
    return {...defaults, ...cfg};
}
```

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -461,7 +460,7 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
     /**
      * How to display the static and contextual actions.
      */
-    @Input() actionDisplayConfig: ActionDisplayConfig = getCopyOfActionDisplayConfig(DEFAULT_ACTION_DISPLAY_CONFIG);
+    @Input() actionDisplayConfig: ActionDisplayConfig = getDefaultActionDisplayConfig();
```
juanmendes, line 463: If we left this null, wouldn't the default from the action menu already apply?

PR #168, opened 2020-08-10
Fixed i18n issues

ryan-bradford:
# Description

Fixes three issues with the i18n library:

1. The translateAsync method would emit an untranslated string prior to emitting a translated string while the translations are loading.
2. The fallbackLocale was not used with async translations.
3. The translationPipe was causing memory issues by creating too many subscriptions. 

# Testing

1. Added unit tests to test the first two issues. 
2. Fixed old unit tests that weren't running. 
3. Loaded into VCD UI and tested the app both in english and German
4. Ran the examples app and made sure translations worked in English.

### DIFF
--- a/projects/i18n/src/lib/pipe/translation-pipe.ts
+++ b/projects/i18n/src/lib/pipe/translation-pipe.ts
@@ -4,6 +4,7 @@
  */
 
 import { ChangeDetectorRef, Injectable, Pipe, PipeTransform } from '@angular/core';
+import { take } from 'rxjs/operators';
 import { TranslationService } from '../service/translation-service';
 
 /**
@@ -35,7 +36,7 @@ export class TranslationPipe implements PipeTransform {
             return key;
         }
 
-        if (key === this.lastKey && params === this.lastArgs) {
+        if (key === this.lastKey && JSON.stringify(params) === JSON.stringify(this.lastArgs)) {
             return this.value;
         }
 
@@ -46,9 +47,12 @@ export class TranslationPipe implements PipeTransform {
     }
 
     private updateValue(key: string, ...args: any[]): void {
-        this.translate.translateAsync(key, args).subscribe(result => {
-            this.value = result;
-            this.changeDetector.markForCheck();
-        });
+        this.translate
+            .translateAsync(key, args)
+            .pipe(take(1))
+            .subscribe(result => {
+                this.value = result;
+                this.changeDetector.markForCheck();
+            });
     }
 }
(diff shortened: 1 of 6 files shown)


projects/i18n/src/lib/pipe/translation-pipe.ts
```diff
@@ -46,9 +47,12 @@ export class TranslationPipe implements PipeTransform {
     }
 
     private updateValue(key: string, ...args: any[]): void {
-        this.translate.translateAsync(key, args).subscribe(result => {
-            this.value = result;
-            this.changeDetector.markForCheck();
-        });
+        this.translate
```
juanmendes, line 50: Calling `toPromise()` seems like a more straight forward way of getting values from an observable without holding on to a subscription.

```ts
const result = await this.translate.translateAsync(key, args).toPromise();
this.value = result;
this.changeDetector.markForCheck();
```

projects/i18n/src/lib/service/message-format-translation-service.spec.ts
```diff
@@ -81,33 +82,58 @@ describe('MessageFormatTranslationService', () => {
         expect(translationService.translate('multiple.params', [{ count: 2, name: 'World' }])).toBe(`many Worlds`);
     });
 
-    it('translates the string via the translation loader via combined translation', () => {
+    it('translates the string via the translation loader via combined translation', done => {
```
juanmendes, line 85: Why use the done callback? That should be used as a last resource. Can't we just make it async and use await?

juanmendes: Weren't there other tests that did the same? We should have them all avoid done since that's the old way of doing things

ryan-bradford: Removed all the usages of `done` from the i18n library. The only place where they still are is in the data exporter specs and I don't think there's a (simple) way to rewrite these without using done.

projects/i18n/src/lib/service/message-format-translation-service.spec.ts
```diff
@@ -81,34 +82,68 @@ describe('MessageFormatTranslationService', () => {
...
             new BehaviorSubject({
                 hi: 'hello',
-            })
+            }).pipe(delay(100))
         );
         const translationService = new MessageFormatTranslationService('en', 'en', loader);
         translationService.registerTranslations();
-        translationService.translateAsync('hi').subscribe(result => {
-            expect(result).toEqual('hello');
+        await translationService
```
juanmendes, line 115: Shouldn't this be

```
const result = await translationService
            .translateAsync('hi')
            .pipe(take(1))
            .toPromise();
expect(result).toEqual('hello');
 expect(translationService.translate('hi')).toEqual('hello');

            expect(result).toEqual('hello');	            .translateAsync('hi')
            .pipe(take(1))
            .toPromise()

PR #173, opened 2020-08-13
Spotlight search improvements

irahov:
# Core component changes:
- use `section` instead of `div`
- use `h5` instead of `div`
- scroll to selected item to ensure it is always visible
- always display section title if such has been provided
      Up to now if there was just one search provider, i.e. one section
      its title was not shown. This leads to confusion when providers
      are added and removed on the fly. In that case one section may apear
      without a title and at the next moment when there is another provider
      this section will have a title.
- add `role="button"` for better ARIA support
- add DOM util class, currently with `scrollToElement` functionality

# Example changes
- new example for a simgle section with no title and many items
- create instance of mouse trap within the constructor to eliminate the
      usage of change detector
- refactor code to make it easier to follow

### DIFF
--- a/projects/components/src/utils/dom-util.spec.ts
+++ b/projects/components/src/utils/dom-util.spec.ts
@@ -1,0 +1,146 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { Component, ElementRef } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { WidgetFinder, WidgetObject } from '../utils/test/widget-object';
+import { DomUtil } from './dom-util';
+
+describe('DomUtil', () => {
+    describe('scrollToElement', () => {
+        // scrollToElement uses smooth scrolling, so we should give the browser some time to do it
+        const SMOOTH_SCROLLING_TIMEOUT = 100;
+
+        interface Test {
+            finder: WidgetFinder<TestHostComponent>;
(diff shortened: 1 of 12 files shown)


projects/components/src/utils/dom-util.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+    public static getHiddenElementCss(): string {
+        return `.element-${TestHostComponent.ELEMENTS_COUNT}`;
+    }
+    public static getVisibleElementCss(): string {
+        return `.element-3`;
+    }
+}
+
+export class TestHostWidgetObject extends WidgetObject<TestHostComponent> {
+    static tagName = `section`;
```
juanmendes, line 126: 👍 I always thought about this, we should use template strings everywhere instead of '' to discourage string concatenation

projects/examples/src/components/spotlight-search/spotlight-search-example.component.ts
```diff
@@ -24,53 +22,44 @@ export class SpotlightSearchExampleComponent implements OnInit, OnDestroy {
...
         private fb: FormBuilder,
         private actionsSearchProvider: ActionsSearchProvider,
-        private spotlightSearchService: SpotlightSearchService,
-        private changeDetectorRef: ChangeDetectorRef
+        private spotlightSearchService: SpotlightSearchService
     ) {
+        // Create an instance of mouse trap within the constructor so that any bount shortcut event handler
+        // would be executed within the angular zone
+        this.mousetrap = new Mousetrap();
+        // For this example we'd like mouse trap to rum always
```
juanmendes, line 35: nit: " For this example, we'd like mouse trap to always run".

 Doesn't that mean we should have an API to do this?

PR #175, opened 2020-08-14
Add directive to change the position of dropdown popover dynamically

ps37:
# Why?
Claritys dropdown does not change its position if it's being clipped and users have to scroll to access the fill dropdown in such cases.
This directive shifts the position of the dropdown popover by checking if a popover is being clipped and tries its best to open it in a direction that's less probable to get clipped based on the available space.

# Testing Done
dynamicDropdownPosition_cc


projects/components/src/dropdown/dynamic-dropdown-position.directive.ts
```diff
@@ -0,0 +1,173 @@
...
+            this.dropdownMenuElement = this.dropdownTriggerElement.querySelector('clr-dropdown-menu');
+            if (this.dropdownMenuElement) {
+                // Recalculate the dropdown position on open
+                this.resetPosition(this.dropdownMenuElement, this.positionTop, this.positionLeft);
+            }
+            try {
+                // Internal API
+                (dropdown as any).popoverInstance.removeScrollEventListeners();
+            } catch (e) {
+                console.warn('Clarity has changed ClrDropdownMenu internal API. Please fix me!');
```
ryan-bradford, line 74: Should we make this a throw error so our tests will fail with a clear reason when we try and bump Clarity if they change the API?

ttsvetko:
Just a quick question:
- i see in the attached gif that power actions are missing - see attached screen-shot from the gif. Is this intentional or a bug?

Screenshot 2020-08-18 at 16 34 35

projects/components/src/dropdown/dynamic-dropdown-position.directive.ts
ps37: I agree

ps37: As for the concerns, that is how it is being done right now in the VCD ui application for quite some time now and no concerns reported yet

ps37:
> Just a quick question:
> 
> * i see in the attached gif that power actions are missing - see attached screen-shot from the gif. Is this intentional or a bug?
> 
> Screenshot 2020-08-18 at 16 34 35

@ttsvetko As you have suggested to me, it looks like this problem is because the content visible area, in this case, is NOT the scroll-div-container that lists cards. And if we have enough space to show the whole menu we should do it, not to hide it.

projects/components/src/dropdown/dynamic-dropdown-position.directive.ts
```diff
@@ -0,0 +1,195 @@
...
+    constructor(
+        private elRef: ElementRef,
+        private renderer: Renderer2,
+        @Inject(DOCUMENT) private document: Document,
+        private dropDownBtn: ClrDropdown,
+        @Optional() @SkipSelf() private parentDropdown: DynamicDropdownPositionDirective
+    ) {}
+
+    ngAfterContentInit(): void {
+        this.isInsideModal = this.document.body.classList.contains(NO_SCROLLING_CLASSNAME);
```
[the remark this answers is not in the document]
ps37: As per the discussion,
- As this is how it has been working in the VCD UI for a while and
- As it is close to the review, we are going to leave it as is and I will file an issue to make the following enhancement

https://github.com/vmware/vmware-cloud-director-ui-components/issues/180

projects/components/src/dropdown/dynamic-dropdown-position.directive.ts
```diff
@@ -143,34 +141,34 @@ export class DynamicDropdownPositionDirective implements AfterContentInit {
         const contentAreaRect = this.contentAreaElement.getBoundingClientRect();
         // If the dropdown is getting clipped on the left and there is enough place on the right, shift the dropdown to right
         if (
-            dropdownMenuRect.left + MENU_BUFFER_SPACE <= contentAreaRect.left &&
-            dropdownTriggerRect.right + dropdownMenuRect.width <= contentAreaRect.right
+            this.dropdownMenuRect.left + MENU_BUFFER_SPACE <= contentAreaRect.left &&
```
juanmendes, line 144: When you're using a reference this many times, you should create a local to make the code look less busy. And it would have saved you time

```const {dropdownMenuRect} = this;```

PR #179, opened 2020-08-19
Change the action menu and dropdown components according to the feedback

ps37:
## What?
Following are the changes made:
- Remove the first level of featured actions dropdown from the action menu
- Add a property in the dropdown item interface to indicate when to add a separator above a dropdown item

## Testing done
- On the vms and vapp screens, making sure that action dropdowns don't have the first level of featured actons dropdown separate from "All actions". Instead, all the actions are displayed in a single dropdown with grouped actions.
- Grouping is as intended and the separators are added on top of the actions that required them.

### Screenshots
Screen Shot 2020-08-19 at 3 17 16 PM
Screen Shot 2020-08-19 at 3 17 38 PM


projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -88,6 +88,10 @@ export interface ActionItem<R, T> {
      * To mark if the {@link #ActionItem.textKey} has to be translated or not
      */
     isTranslatable?: boolean;
+    /**
+     * To add separators between groups of action
```
juanmendes, line 92: should be "between groups of action`s`"

PR #186, opened 2020-08-26
Performance improvements of action menu component

ps37:
## Why?
Getters with looping and recursive logic over nested action lists are being used in the HTML templates. Because of this, On any angular 'digest' cycle these getter functions will be called over and over again. Internally these will call the logic which iterates through an array and so on and so forth

## What?
As part of this review, the following are the changes made:
- Address Ivos review comments on https://github.com/vmware/vmware-cloud-director-ui-components/pull/135
- Change the above-mentioned getters into class properties

## Testing done:
Copy pasted the components package with the above changes in vcd_ui code base
Verified that actions are displayed as before on the following screems
- All screens of vapps and vms
- Users list screen
- Networks list screen

actionMenuPerfImprovements


projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -33,46 +33,38 @@ export function getDefaultActionDisplayConfig(cfg: ActionDisplayConfig = {}): Ac
...
 export class ActionMenuComponent<R, T> {
-    private _actions: ActionItem<R, T>[] = [];
     /**
      * List of actions containing both static and contextual that are given by the calling component
      */
     @Input() set actions(actions: ActionItem<R, T>[]) {
         if (!actions) {
             return;
         }
-
```
juanmendes, line 44: We recently chatted about not modifying data. Can we avoid modifying this data?

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -145,22 +186,59 @@ export class ActionMenuComponent<R, T> {
...
+        const copyOfActions = getDeepCopyOfActionItems(actions);
         if (!this.calculateActionsAvailability) {
-            return actions;
+            return copyOfActions;
         }
-        return actions.filter(action => {
+        return copyOfActions.filter(action => {
             const isActionAvailable = this.isActionAvailable(action);
-            if (isActionAvailable && action.children && action.children.length) {
+            if (isActionAvailable && !!action.children?.length) {
```
juanmendes, line 227: Is the `!!`needed? I don't think so. A filter callback should never ever modify data. Please re-work.

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -627,6 +614,18 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
         this.changeDetectorRef.detectChanges();
     }
 
+    private getHasStaticActions(): boolean {
```
juanmendes, line 617: `getHasStatic`? What was wrong with its previous name?

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -88,14 +80,55 @@ export class ActionMenuComponent<R, T> {
...
+     * the contextual actions dropdown in the inline actions container
+     */
+    shouldDisplayContextualActionsDropdownInline = false;
+
+    private _actionDisplayConfig: ActionDisplayConfig = getDefaultActionDisplayConfig();
+
+    /**
+     * Used for calculating the available actions. So, updating the property should update available actions in the action lists
+     */
+    @Input() set calculateActionsAvailability(val: boolean) {
```
juanmendes, line 114: A setter with a verb? I can't make sense of it. It's better not to say how it's used, you should explain its meaning. Why does the parent set this flag, this seems internal?

irahov: I agree that setting `calculateActionsAvailability` from outside looks awkward - what is the case when actions availability should not be calculated?
In addition if I use `[calculateActionsAvailability]="false"` I would expect they are not calculated, but looking in the code the `this.updateActionLists();` will be called

projects/components/src/action-menu/action-menu.component.ts
irahov, line 227: nit: 
I find it easier to read if it is:
`isActionAvailable && action.children?.length > 0`
I mean I read it as English language sentence.

irahov (approved):
Please make sure that after the improvement the action availability is calculated correctly

projects/components/src/action-menu/action-menu.component.ts
ps37: Added a separate Jira task to fix this in a different PR in Andromeda release https://jira.eng.vmware.com/browse/VDUCC-323

juanmendes (approved):

PR #190, opened 2020-08-31
Enable projecting content when there is no search

irahov:
# Description
There is quite much empty space when there is no search which
to be utilized in one way or another.

One example is to display some helpful information on how the spotlight search can be used.

# Testing done:
By using the spotlight search examples add text which is shown when there is
search, i.e. the search input is empty.

# Screenshots
no-search text is shown when there is no search
image

no-search text is hidden when there is a search
image

### DIFF
--- a/projects/examples/src/components/spotlight-search/spotlight-search-example.component.scss
+++ b/projects/examples/src/components/spotlight-search/spotlight-search-example.component.scss
@@ -1,0 +1,6 @@
+.no-search {
+    height: 100%;
+    display: flex;
+    align-items: center;
+    text-align: center;
+}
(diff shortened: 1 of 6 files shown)


projects/examples/src/components/spotlight-search/spotlight-search-example.component.scss
```diff
@@ -0,0 +1,3 @@
+.no-search {
+    margin-top: 1.5rem;
```
juanmendes, line 2: Instead of margin-top, can we make it center aligned vertically?

irahov:
As discussed offline with UX this functionality should not be applied.

PR #191, opened 2020-08-31
Improve API for registering components

juanmendes:
## Features

* Registration no longer returns an ID, you just pass the provider to be unregistered but that is now hidden behind QuickSearchRegistrar
* Providers can auto unregister when their calling components are destroyed bu using the QuickSearchRegistrar
* Providers provide the section name when they receive it,
  no need to re-register after you figure out the section name
* Order of provider within quick search is also determined by the provider, not a parameter to registration

### Features (probably should have been a different review but would not happen before release

* Rename to quick search instead of using Apple's spotlight
* Update prettier and tslint so we can use null-ish coalescing feature of TS

## Testing Done
* Ran examples container for spotlight example making sure providers are destroyed when leaving a page, that is, there are no duplicate providers as you navigate between examples

quick-search


projects/components/src/quick-search/quick-search.component.ts
```diff
@@ -295,6 +299,11 @@ export class SpotlightSearchComponent {
     showSectionTitle(searchSection: SearchSection): boolean {
         // In order to show a section title there should be more than one sections
         // and the current section should either be loading data or have results
-        return searchSection.section && (searchSection.isLoading || searchSection.results.length > 0);
+        console.log(
```
juanmendes, line 302: Delete console.log

tsconfig.json
```diff
@@ -20,7 +20,7 @@
     ],
     "paths": {
       "@vcd/ui-components": [
-        "dist/components"
+        "projects/components/src"
```
irahov, line 23: I like this is in, I hope no side effects for the official github build

irahov (approved):

projects/components/src/quick-search/quick-search.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    providedIn: 'root',
+})
+export class QuickSearchService {
+    registrations: QuickSearchProvider[] = [];
+
+    /**
+     * Register a search provider
+     * @param provider The search provider {@link QuickSearchProvider}
+     */
+    public registerProvider(provider: QuickSearchProvider): void {
```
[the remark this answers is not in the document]
juanmendes: It's because unregistering can kind of fail if the provider is not registered. Since you added unit tests to it, I just kept it working, it doesn't seem wrong

PR #192, opened 2020-08-31
Two way binding for selection

ryan-bradford:
# Description

The grid now has two generics:
B  - an object that contains only the single property that identifies this record. So for example a { href: string }. This is the same property that your trackBy MUST return.
R - the record itself. This type extends B.
The property datagridSelection is now a two way binding that accepts/returns a (R | B)[]. If you want to set some list of preselected items, create a list of B's and bind to datagridSelection . datagridSelection will then continue to output a type B for all unloaded entities, but will output a full type R for any entity that has been loaded to the grid.

This grid also has an input preserveSelection that defaults to false. preserveSelection must be set to true for selection to be maintained across multiple pages/filters.

### Not a breaking change

The default behavior is the same as it was before. It is only when you set preserveSelection to true does the behavior change.

# Testing

1. Viewed the examples app and saw selection work across many pages.
2. Added some unit tests
3. Loaded into VCD_UI and saw selection continue to work.

select-change

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.ts
+++ b/projects/components/src/datagrid/datagrid.component.ts
@@ -20,11 +20,10 @@ import {
     ViewChildren,
 } from '@angular/core';
 import { ClrDatagrid, ClrDatagridPagination, ClrDatagridStateInterface } from '@clr/angular';
+import { SelectionType } from '@clr/angular/data/datagrid/enums/selection-type';
 import { LazyString, TranslationService } from '@vcd/i18n';
 import { Observable } from 'rxjs';
-import {
-    ActionMenuComponent,
-} from '../action-menu/action-menu.component';
+import { ActionMenuComponent } from '../action-menu/action-menu.component';
 import { ActivityReporter } from '../common/activity-reporter';
 import {
     ActionDisplayConfig,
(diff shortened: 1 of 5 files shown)


projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -423,20 +417,36 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
     /**
      * Returns the items selected in the VCD datagrid.
      */
-    get datagridSelection(): R[] {
+    @Input()
+    get datagridSelection(): (B | R)[] {
```
juanmendes, line 421: `B | R` sounds wrong if `R` extends `B`. Shouldn't it just be `B`? And we can specify that callers must cast and test if if they want to use `R` since we can't guarantee it... 

If this is really required, we should define a type instead of copy pasting `(R | B)[]`

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -456,6 +466,43 @@ export class DatagridComponent<R> implements OnInit, AfterViewInit, OnDestroy {
         return this.contextualActionPosition && this.contextualActionPosition === ContextualActionPosition.ROW;
     }
 
+    /**
+     * The {@link maxFeaturedActionsOnRow} value depends in the contextual featured actions config which belongs to the
```
juanmendes, line 470: If you're pointing to a property of the current class, the correct syntax is `{@link #maxFeaturedActionsOnRow}`

PR #193, opened 2020-08-31
Display fieldName on 'export all columns' of data-exporter

yumengwu:
Display `fieldName` on 'export all columns' of data exporter modal when there is no `displayName`.

**Why this change made?**
After making `displayName` an optional property, for those export columns only having `fieldName`, the column selection on data exporter will display empty text. This change could make it display fieldName.

Test Done:
 - Add a test case to verify `clr-dropdown-menu` and `column-container` display `fieldName` when there is no `displayName`.
 - manually: after opening the data exporter modal and unchecking 'export all columns', the export columns only having `fieldName` display `fieldName` now.

Before this change:

After this change:
Screen Shot 2020-08-31 at 3 37 47 PM
Screen Shot 2020-08-31 at 3 37 59 PM

### DIFF
--- a/projects/components/src/data-exporter/data-exporter.component.html
+++ b/projects/components/src/data-exporter/data-exporter.component.html
@@ -65,7 +65,9 @@
                                         <clr-checkbox-wrapper class="column-checkbox">
                                             <input type="checkbox" clrCheckbox [formControlName]="col.fieldName" />
                                             <label>{{
-                                                this.friendlyFieldsControl.value ? col.displayName : col.fieldName
+                                                this.friendlyFieldsControl.value
+                                                    ? col.displayName || col.fieldName
+                                                    : col.fieldName
                                             }}</label>
                                         </clr-checkbox-wrapper>
                                     </li>
@@ -77,7 +79,7 @@
             </ul>
             <div [ngStyle]="{ visibility: shouldShowBubbles ? 'visible' : 'hidden' }" class="column-container">
                 <span class="label column-label" *ngFor="let column of selectedColumns">
-                    {{ this.friendlyFieldsControl.value ? column.displayName : column.fieldName }}
+                    {{ this.friendlyFieldsControl.value ? column.displayName || column.fieldName : column.fieldName }}
                     <clr-icon shape="close" class="badge badge-info" (click)="selectColumn(column, false)"> </clr-icon>
                 </span>
             </div>
--- a/projects/components/src/data-exporter/data-exporter.component.spec.ts
+++ b/projects/components/src/data-exporter/data-exporter.component.spec.ts
@@ -48,16 +48,32 @@ describe('DataExporterColumnsWithoutDisplayName', () => {
                 ...TestData.exportDataWithoutDisplayName.map(row => Object.values(row)),
             ];
             const csvString = downloadService.createCsv(exportData);
-            expect(downloadService.downloadCsvFile).toHaveBeenCalledWith(
-                csvString,
-                exporter.component.fileName
-            );
+            expect(downloadService.downloadCsvFile).toHaveBeenCalledWith(csvString, exporter.component.fileName);
             this.finder.detectChanges();
             expect(this.finder.hostComponent.dataExporterOpen).toBe(false);
             done();
         });
 
         exporter.clickExport();
+    });
+
+    it('displays field name when there is no display name', function(this: TestExporterColumnsWithoutDisplayNameFinder, done): void {
+        const exporter = this.finder.find(DataExporterWidgetObject);
+        let dropdownMenu = document.querySelector('clr-dropdown-menu');
+        expect(dropdownMenu).toBeNull();
+        exporter.component.selectAllControl.setValue(false);
+        this.finder.detectChanges();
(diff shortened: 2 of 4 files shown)


juanmendes:
Please create a unit test that shows we default display name to  field name

projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -65,7 +65,9 @@ <h3 class="modal-title">{{ dialogHeader | lazyString }}</h3>
                                         <clr-checkbox-wrapper class="column-checkbox">
                                             <input type="checkbox" clrCheckbox [formControlName]="col.fieldName" />
                                             <label>{{
-                                                this.friendlyFieldsControl.value ? col.displayName : col.fieldName
+                                                this.friendlyFieldsControl.value
```
juanmendes, line 68: Shouldn't this just be `col.displayName || col.fieldName` ?

ryan-bradford: No, because in the case that this.friendlyFieldsControl.value is true, we want to display the `col.displayName` with a backup of `col.fieldName`, but in the case it is false, we only want to display `col.fieldName`

projects/components/src/data-exporter/data-exporter.component.spec.ts
```diff
@@ -48,17 +48,33 @@ describe('DataExporterColumnsWithoutDisplayName', () => {
...
+    it('displays field name when there is no display name', function(this: TestExporterColumnsWithoutDisplayNameFinder, done): void {
+        const exporter = this.finder.find(DataExporterWidgetObject);
+        let dropdownMenu = document.querySelector('clr-dropdown-menu');
+        expect(dropdownMenu).toBeNull();
+        exporter.component.selectAllControl.setValue(false);
+        this.finder.detectChanges();
+        dropdownMenu = document.querySelector('clr-dropdown-menu');
+        const selections = dropdownMenu.querySelectorAll('li');
+        expect(selections.length).toEqual(2);
+        expect(selections[0].querySelector('label').innerText.trim()).toEqual('col1');
```
juanmendes, line 69: We don't make queries from jasmine, all queries should be hidden behind the widget wrapper, you are duplicating code from the widget wrapper

projects/components/src/data-exporter/data-exporter.component.spec.ts
```diff
@@ -332,6 +348,7 @@ class TestHostComponent {
 @Component({
     template: `
         <vcd-data-exporter
+            id="dataExporterWithoutDisplayName"
```
juanmendes, line 351: Why this ID? We should not hard code IDs

projects/components/src/data-exporter/data-exporter.component.ts
```diff
@@ -347,11 +360,18 @@ export class DataExporterComponent implements OnInit, OnDestroy {
...
+            const exportColumn = this.fieldNameMap.get(fieldName);
+            return exportColumn.displayName || exportColumn.fieldName;
+        } else {
+            return fieldName;
         }
-        return fieldName;
+    }
+
+    private updateFieldNameMap(cols: ExportColumn[]): void {
+        this.fieldNameMap.clear();
```
juanmendes, line 372: A better way is not to mutate objects

```
this.fieldNameMap = new Map(cols.map( col => [col.fieldName, column]))

```

projects/components/src/data-exporter/data-exporter.component.ts
```diff
@@ -347,11 +360,18 @@ export class DataExporterComponent implements OnInit, OnDestroy {
     }
 
     private getDisplayNameForField(fieldName: string): string {
-        for (const column of this.columns) {
-            if (column.fieldName === fieldName) {
-                return column.displayName || column.fieldName;
-            }
+        if (this.fieldNameMap.has(fieldName)) {
```
juanmendes, line 363: If your if clause has a return, there's no need for an else. This practice minimizes indentation.

When would this be called with a `fieldName` that doesn't exist? I don't see a need for the else, I think we're not getting coverage for this, I don't think it's possible this is ever hit unless we write a bug.

projects/components/src/data-exporter/data-exporter.component.html
```diff
@@ -77,7 +79,7 @@ <h3 class="modal-title">{{ dialogHeader | lazyString }}</h3>
             </ul>
             <div [ngStyle]="{ visibility: shouldShowBubbles ? 'visible' : 'hidden' }" class="column-container">
                 <span class="label column-label" *ngFor="let column of selectedColumns">
-                    {{ this.friendlyFieldsControl.value ? column.displayName : column.fieldName }}
+                    {{ this.friendlyFieldsControl.value ? column.displayName || column.fieldName : column.fieldName }}
```
juanmendes, line 82: The point of my last request was so we wouldn't have to duplicate the defaulting here, you should be using the function that gets the display name.

projects/components/src/data-exporter/data-exporter.component.html
juanmendes: Should be calling getDisplay name instead of duplicating the defaulting

PR #194, opened 2020-09-02
Display total number of results per search provider.

irahov:
# Description
The search result of a provider can return a subsection of the total number
of items that match the criteria. Such an example is a backend query. The ideal
behavior is to be able to load the next page but we are still not there to implement this.

The current solution for this scenario is to show the number of the found items,
the total number of items and a warning message to the user.

# Changes
The changes include:
- modify the QuickSearchResult to contain items, page, pageSize and total
- update the search component to
  -- show the partial results found
  -- display a warning message if there are more items
- add new SearchProvider for the sync-async example

# Screenshots
image
image

### DIFF
--- a/projects/components/src/assets/resources/en.properties
+++ b/projects/components/src/assets/resources/en.properties
@@ -21,6 +21,8 @@ vcd.cc.friendly.names.hint=User friendly column headers or object field names?
 vcd.cc.export.all=Export All Columns
 vcd.cc.exporter.downloading=Downloading...
 vcd.cc.exporter.writing=Writing File...
+vcd.cc.quickSearch.partialResultNotation=({lastItem} of {totalItems})
+vcd.cc.quickSearch.refineQuery=The search shows a maximum of {max} results. Refine your query to improve the accuracy.
 
 vcd.cc.units.bytes.B=B
 vcd.cc.units.bytes.KB=KB
--- a/projects/components/src/quick-search/quick-search-result.ts
+++ b/projects/components/src/quick-search/quick-search-result.ts
@@ -4,11 +4,36 @@
  */
 
 /**
- * The interface a spotlight search result item should implement
+ * The interface of a quick search result
  */
-export interface QuickSearchResult {
+export interface QuickSearchResults {
     /**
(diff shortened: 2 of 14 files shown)


projects/components/src/quick-search/quick-search.component.html
```diff
@@ -23,6 +23,9 @@
             >
                 <h5 *ngIf="showSectionTitle(searchSection)" class="search-result-section-title">
                     {{ searchSection.provider.sectionName }}
+                    <span *ngIf="partialResult(searchSection); let partial" class="partial-result">
+                        {{ 'vcd.cc.quickSearch.pagination' | translate: partial }}
```
irahov, line 27: What about making `'vcd.cc.quickSearch.pagination'` to be an input property (message key) of the provider?

projects/components/src/assets/resources/en.properties
```diff
@@ -22,6 +22,10 @@ vcd.cc.export.all=Export All Columns
 vcd.cc.exporter.downloading=Downloading...
 vcd.cc.exporter.writing=Writing File...
 
+#Quick Search
```
juanmendes, line 25: Do we really need this header? All the keys already say quicksearch

If so, please add a space between `#` and `Quick`

projects/components/src/quick-search/quick-search.component.html
juanmendes: I think we can add that later since we don't really need it now

projects/components/src/quick-search/quick-search-result.ts
```diff
@@ -7,6 +7,31 @@
  * The interface a spotlight search result item should implement
  */
 export interface QuickSearchResult {
+    /**
```
juanmendes, line 10: I don't understand why we need two interfaces with a very similar comment

> The interface a quick search result item should implement. 

It seems like the first one should be `QuickSearchResults` since it's a list of results and the second one `QuickSearchResultItem` should be private, users of this class should only use `QuickSearchResultsType` (adding the s since it's a list)

juanmendes (approved):

PR #198, opened 2020-09-08
Quick search improvements

irahov:
A number of changes:

Add capability to project content into the quick search
Always show section name in the quick search
Add data property to the search providers
Fix bug that manually selected item is not preserved in async case

Please take a look at the individual comments


projects/components/src/quick-search/quick-search.component.ts
```diff
@@ -198,7 +200,9 @@ export class QuickSearchComponent {
             }
             searchSection.result = searchResult;
             searchSection.isLoading = false;
-            this.selectFirst(true);
+            if (this.selectedItem == null) {
```
juanmendes, line 203: We usually prefer `=== null` unless you are trying to specifically test for undefined also

projects/components/src/quick-search/quick-search.service.spec.ts
```diff
@@ -4,11 +4,15 @@
...
-import { QuickSearchProvider } from './quick-search.provider';
+import { QuickSearchProviderDefaults } from './quick-search.provider';
 import { QuickSearchService } from './quick-search.service';
 
-class SimpleSearchProvider implements QuickSearchProvider {
-    constructor(public sectionName: string = '', public order: number = -1) {}
+class SimpleSearchProvider extends QuickSearchProviderDefaults {
+    constructor(sectionName: string = '', order: number = -1) {
+        super();
+        this.sectionName = sectionName;
```
juanmendes, line 13: Why was this changed, previous syntax is shorter

projects/examples/src/components/quick-search/quick-search-sync-async.example.component.html
```diff
@@ -16,6 +16,21 @@
         description="The text to be shown in the quick search input field when it is empty"
     >
     </vcd-form-input>
+    <vcd-form-checkbox
```
juanmendes, line 19: Ideally, we'd have an example for this feature, not have this embedded in an example that is meant to show async results

juanmendes (approved):

projects/examples/src/components/quick-search/quick-search-sync-async.example.component.html
irahov: reverted

PR #202, opened 2020-09-16
Number with unit input - various fixes

irahov:
# Description

Some of the changes are:
* Add new examples, enhance existing ones
* Enable the component to work with no units
* Remove internal validation but enable external ones only
* Add new validator: NumberWithUnitsFormValidatorsFactory.isInRange
* (in this component only) support not only `{ [key]: true }` error, but also `{ [key]: any }`
  This gives much more flexibility when translating errors 
* Use error style of the input, but show errors outside of the input component
  for 2 reasons:
  a) use the new supported format of errors and
  b) align the error not only under the input element but also under the unit dropdown
* Add initialValueUnit to prevent finding the best match for the initial value
* min/max values of the inner input are calculated according to the selected unit
* remember last real value when switching back and forth from 'unlimited' values
* fix buggy behavior when setting formValues from outside
* fix styling of readonly value
* fix misaligned 'unlimited' value
* fix disabled state not disabling unlimited value

# Testing done
Add new examples and play with them - setting different values, options etc.

### DIFF
--- a/projects/examples/src/components/form-input/number-with-unit-form-input-unitless.example.component.html
+++ b/projects/examples/src/components/form-input/number-with-unit-form-input-unitless.example.component.html
@@ -1,0 +1,56 @@
+These examples show how you can use the <code>&lt;vcd-number-with-unit-form-input&gt;</code> component without units or
+with percent.
+<br />
+The "Warn Level" percent example also disables the "unlimited" functionality.
+<p>
+    You can play with the form below to examine validators, min/max values, unlimited values, readonly & disabled state
+    etc.
+</p>
+<form [formGroup]="formGroup" class="clr-form-horizontal">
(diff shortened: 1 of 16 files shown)


projects/examples/src/components/form-input/number-with-unit-form-input-unitless.example.component.html
```diff
@@ -0,0 +1,42 @@
+<form [formGroup]="formGroup" class="clr-form-horizontal">
```
juanmendes, line 1: Can we add some text to the example giving an overview of the feature we're exemplifying?

juanmendes (approved):

juanmendes:
Please explain the issues being fixed in the description

PR #205, opened 2020-09-17
Remove extension action menu directive and @vcd/sdk dependency

ps37:
# Why?
We created a new service in vcd ui called EntityActionExtensionService(https://gitlab.eng.vmware.com/core-build/vcd_ui/merge_requests/2650) that will take care of the plugin actions, this directive is no longer required.
Because of that the @vcd/sdk dependency is also no longer required.

# Testing done:
Tested the core components examples app is working without any problems


juanmendes (approved):

ps37:
The changes in this PR which involve removing of @vcd/sdk dependency are already pushed to master branch.

PR #210, opened 2020-09-22
Fixes bugs with the number with unit input

ryan-bradford:
# Description

1. Fixes a bug where the findBestUnit method did not work correctly
2. Fixes a bug when on editing an input without an unlimited checkbox, the best unit was not selected. 

# Testing

Loaded into VCD_UI.
1. Created a Vapp from a template to show that best unit selection now works for their memory input. 
2. Edited a Kubernetes policy on a Org VDC to show their validators working with the correct units. 

number-fix
Screen Shot 2020-09-22 at 1 46 01 PM

### DIFF
--- a/projects/components/src/utils/unit/unit.ts
+++ b/projects/components/src/utils/unit/unit.ts
@@ -72,27 +72,27 @@ export abstract class Unit {
 }
 
 /**
- * Finds the bestUnit by trying groups of thousands
+ * Finds the bestUnit by trying groups of the comparison number
  */
-export abstract class ThousandsUnit extends Unit {
+abstract class NumberUnit extends Unit {
     /**
      * Calculates the best unit out of available units to display in UI cell for a given input Unit
      * and value
      *
-     * Best unit is a Unit whose converted value is less than 1000
+     * Best unit is a Unit whose converted value is less than the comparison number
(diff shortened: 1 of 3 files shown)


juanmendes (approved):

projects/components/src/utils/unit/unit.ts
```diff
@@ -102,7 +102,27 @@ export abstract class ThousandsUnit extends Unit {
         }
     }
 
-    abstract getAllUnitTypes(): ThousandsUnit[];
+    abstract getAllUnitTypes(): NumberUnit[];
+
+    protected abstract getComparisonNumber(): number;
```
juanmendes, line 107: Can we rename this to `getGroupingNumber()` and add some comments.

juanmendes (approved):

projects/components/src/utils/unit/unit.ts
irahov: `getGroupingNumber` is not clear to me.
Isn't it related to a `multiplier` ?

ryan-bradford: Yea, I think Ivo's right. How does `getUnitMultiplier` sound?

PR #215, opened 2020-09-25
Fix actions not shown until entity is loaded bug and action search result availability calculation bug

ps37:
## This fixes the following bugs:
- Displaying `?undefined` in the action search result section. This is happening because of not filtering out the `separator` objects from the action list.
- Actions are not shown after the entity is done loading. The search text has to be changed for them to get updated.

## Testing done:
- Verified that the vm and vapp entity details screens of vcd ui don't show errors 1 and 2

actionSeach_bugs1 2

## ToDo:
- Unit tests for newly added code in action search provider
- Example showcasing the added pause and unpause functionality of action search

### DIFF
--- a/projects/components/package.json
+++ b/projects/components/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@vcd/ui-components",
-    "version": "1.0.0-dev.43",
+    "version": "1.0.0-dev.44",
     "dependencies": {
         "@ngx-formly/core": "5.6.1"
     },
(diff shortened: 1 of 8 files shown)


juanmendes:
Please add unit tests, and an example for pause and unpause...

juanmendes:
Approved after unit tests are added

juanmendes (approved):

juanmendes (approved):

PR #221, opened 2020-10-13
Prevent nested menus that are opened on hover from closing when clicked

ps37:
# Problem:
Nested-menus that are opened on hover are being closed for example when they are accidentally clicked upon. This is not a very good UX.

# Solution:
Looking at other dropdowns like... in Mac OS and Microsft outlook, we found out that the nested menus that are opened on hover are not reacting to mouse clicks. So, taking the inspiration from there, I have done something similar by preventing mouse clicks from happening on the nested menu triggers.

# Testing done:
dropdown_notOpenOnClick

### DIFF
--- a/projects/components/src/dropdown/dropdown.component.ts
+++ b/projects/components/src/dropdown/dropdown.component.ts
@@ -4,6 +4,7 @@
  */
 
 import {
+    AfterViewInit,
     Component,
     Host,
     HostListener,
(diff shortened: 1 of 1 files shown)


juanmendes:
When the menu is expanded, focus should go to the first item in the child drop down to keep the behavior similar to what happens when you navigate with the arrows on the keyboard

ps37:
> When the menu is expanded, focus should go to the first item in the child drop down to keep the behavior similar to what happens when you navigate with the arrows on the keyboard

This is being addressed as part of the custom handling logic at https://github.com/vmware/vmware-cloud-director-ui-components/pull/224

juanmendes (approved):

PR #222, opened 2020-10-13
Unit tests for dropdown menu mouse event handlers

ps37:
Adds unit tests to cover the logic that opens and closes drop-down menus using `mouseover` and `mouseout` events

### DIFF
--- a/projects/components/src/dropdown/dropdown.component.spec.ts
+++ b/projects/components/src/dropdown/dropdown.component.spec.ts
@@ -192,7 +192,7 @@ describe('DropdownComponent', () => {
             ).toEqual(false);
         });
     });
-    describe('open and close drop downs,', () => {
+    describe('toggling,', () => {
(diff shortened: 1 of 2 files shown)


projects/components/src/dropdown/dropdown.component.spec.ts
```diff
@@ -145,10 +183,95 @@ describe('DropdownComponent', () => {
...
+                declarations: [TestHostComponent],
+            }).compileComponents();
+
+            this.finder = new WidgetFinder(TestHostComponent);
+            this.finder.detectChanges();
+            this.dropdownWidget = this.finder.find(VcdDropdownWidgetObject);
+            this.dropdownComponent = this.dropdownWidget.component;
+        });
+
+        it('clicking the primary dropdown toggles the primary dropdown menu', function (this: HasVcdDropdown): void {
```
juanmendes, line 214: You should try to make the text read like `it does something` when possible.

```it("toggles the primary dropdown menu when clicked")```

PR #224, opened 2020-10-21
Add custom focus handling logic for dropdown menu

ps37:
## Why this change:
Because the focus handling logic through the keyboard is not working correctly on the vcd-dropdown component. This is happening because the Clarity code is not able to recognize nested Vcd dropdown as content children. So we decided to write our own focus handling logic

## What is the change:
Disable the Claritys focus handling logic and then write our own focus handling logic. A directive to link the menu items and then service for updating the currently focused item. There is only one instance of this per root dropdown.

## Testing done:
customFocusHandling_dropdown

### DIFF
--- a/projects/components/src/dropdown/dropdown-focus-handler.directive.ts
+++ b/projects/components/src/dropdown/dropdown-focus-handler.directive.ts
@@ -160,13 +160,17 @@ export class DropdownFocusHandlerDirective<T> implements AfterViewInit, OnDestro
     private linkMenuItems(): void {
         const menuChildren: Element[] = Array.from(this.clrDropdownMenuEl.children);
         this.menuItems = menuChildren.map((child) => ({
-            element:
-                child.nodeName === BUTTON_NODE_NAME
-                    ? (child as HTMLElement)
-                    : (child.querySelector('clr-dropdown > button') as HTMLElement),
+            element: (child.matches('button') ? child : child.querySelector('clr-dropdown > button')) as HTMLElement,
+            left: this.menuTrigger,
         }));
         this.linkVertical();
-        this.linkMenuToTrigger();
+        // Lint to menu trigger
+        if (this.isRootDropdown) {
+            this.menuTrigger.down = this.menuItems[0];
+        } else {
+            // If there is a parent dropdown, We have to link these menu items to their trigger in the parent dropdown
+            this.menuTrigger.right = this.menuItems[0];
+        }
     }
 
     private get rootMenuTrigger(): MenuItem {
@@ -186,16 +190,6 @@ export class DropdownFocusHandlerDirective<T> implements AfterViewInit, OnDestro
             this.hostVcdDropdown.clrDropdown.toggleService.open = false;
         };
         return menuTrigger;
-    }
-
-    private linkMenuToTrigger(): void {
-        if (this.isRootDropdown) {
-            this.menuTrigger.down = this.menuItems[0];
-            return;
-        }
-        // If there is a parent dropdown, We have to link these menu items to their trigger in the parent dropdown
-        this.menuTrigger.right = this.menuItems[0];
-        this.menuItems.forEach((item) => (item.left = this.menuTrigger));
     }
 
     private linkVertical(): void {
--- a/projects/components/src/dropdown/dropdown-focus-handler.directive.ts
+++ b/projects/components/src/dropdown/dropdown-focus-handler.directive.ts
@@ -1,0 +1,165 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { AfterViewInit, Directive, Host, OnDestroy, Optional, SkipSelf } from '@angular/core';
+import { ClrPopoverToggleService } from '@clr/angular';
+import { Subscription } from 'rxjs';
+import { SubscriptionTracker } from '../common/subscription';
+import { DropdownFocusHandlerService } from './dropdown-focus-handler.service';
+import { DropdownComponent } from './dropdown.component';
+
+const BUTTON_NODE_NAME = 'BUTTON';
+
+/**
+ * Arrow keys directions
+ */
+export enum Direction {
+    UP = 'up',
+    DOWN = 'down',
+    LEFT = 'left',
+    RIGHT = 'right',
+}
+
+/**
+ * Object wrapping the focusable HTML elements of dropdown menu with the neighbors in 4 directions of each menu item.
+ */
+export interface MenuItem {
+    element: HTMLElement;
+    up?: MenuItem;
+    down?: MenuItem;
+    left?: MenuItem;
+    right?: MenuItem;
+    closeMenu?: (event: Event) => void;
+}
+
+/**
+ * Added on to vcd-dropdown component to vertically link menu items of the dropdown on which this directive is added. It also, then links
+ * the vertically linked menu items horizontally to their menu trigger. It then uses the {@link DropdownFocusHandlerService} to move the
+ * DOM focus between the menu items.
+ *
+ * @Example:
+ * <vcd-dropdown
+ *        vcdDropdownFocusHandler
+ *        [items]="contextualActions"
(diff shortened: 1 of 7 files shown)


projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    private moveFocusToFirstItem(): void {
+        this.focusHandlerService.moveFocusTo(this.menuTrigger);
+        // After the menu is opened, the ClrDropdownMenu wouldn't be attached to the DOM until the next change detection cycle.
+        // So setTimeout is the only way to wait for the ClrDropdownMenu to be ready to move focus to first item.
+        this.timeoutId = window.setTimeout(() => {
+            if (this.parentVcdDropdown) {
+                this.focusHandlerService.moveFocus(Direction.RIGHT);
+            } else {
+                this.focusHandlerService.moveFocus(Direction.DOWN);
```
ryan-bradford, line 104: Is this method called when an item is clicked? Does dropdownMenuUpdated emit an item on click? What happens if we have A menu with three items, we have the second one selected, and we click the top one:
A
B (focused)
C
 
It feels like the way this is implemented it would turn into 
A
B
C (focused)

but it should turn into:
A (focused)
B
C

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+        @Host() private hostVcdDropdown: DropdownComponent<T>,
+        private focusHandlerService: DropdownFocusHandlerService
+    ) {}
+
+    /**
+     * List of focusable menu items with their neighbors in 4 directions.
+     */
+    menuItems: MenuItem[];
+
+    private menuTrigger: MenuItem;
```
ryan-bradford, line 66: Can we either put a comment on this property or rename it to be more clear? I'm not totally sure what it represents

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * After a dropdown menu is opened, it creates {@link MenuItem} for each of the menu items along with their trigger menu item and links
+     * them. It also then moves the focus to first item in the opened menu list.
+     */
+    ngAfterViewInit(): void {
+        this.subscriptionTracker.subscribe(this.hostVcdDropdown.dropdownMenuUpdated, (updated) => {
+            if (!updated) {
+                return;
+            }
+            this.createMenuTrigger();
```
ryan-bradford, line 80: Feels like the process here should be:
1. Update the graph of nodes which has a few substeps
   1. Get current item, link to this dropdown
   2. Link all items in this dropdown to current item
   3. Link all items in this dropdown to each other
2. Move focus to first item

Do you think we should combine the first two methods into one with appropriate substeps to something like `updateMenuGraph()`? Also, can the call to `this.registerRootMenuContainer();` be moved to the constructor or ngOnInit?

projects/components/src/dropdown/dropdown.component.ts
```diff
@@ -5,24 +5,43 @@
...
+    renderer: Renderer2,
+    existing: DropdownFocusHandlerService
+): DropdownFocusHandlerService {
+    return existing || new DropdownFocusHandlerService(renderer);
+}
+
+export const DROPDOWN_FOCUS_HANDLER_PROVIDER: Provider = {
+    provide: DropdownFocusHandlerService,
+    useFactory: dropdownFocusHandlerServiceFactory,
+    deps: [Renderer2, [new Optional(), new SkipSelf(), DropdownFocusHandlerService]],
```
ryan-bradford, line 42: Does this generate a singelton for all dropdowns even if there are many on the same screen? Or is it smart enough to generate a different one per dropdown but not nested dropdowns?

projects/components/src/dropdown/dropdown.component.ts
```diff
@@ -88,7 +108,37 @@ export class DropdownComponent<T extends DropdownItem<T>> {
...
+
+    constructor(@Optional() @SkipSelf() private parentVcdDropdown: DropdownComponent<T>) {}
+    @ViewChild(ClrDropdownMenu)
+    set clrDropdownMenu(val: ClrDropdownMenu) {
+        this._clrDropdownMenu = val;
+        if (!val) {
+            return;
+        }
+        // Disable Claritys focus handling logic
+        console.log(`${this._dropdownTriggerEl.innerText}s menu is opened. ${this.clrDropdown.toggleService.open}`);
```
ryan-bradford, line 121: Remove console.logs

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+                    ? (child as HTMLElement)
+                    : (child.querySelector('clr-dropdown > button') as HTMLElement),
+        }));
+
+        this.linkVertical();
+
+        this.linkMenuToTrigger();
+    }
+
+    private createMenuTrigger(): void {
```
ryan-bradford, line 124: Can we rename this to something like `findCurrentSelectedItem()`?

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+            this.registerRootMenuContainer();
+            this.moveFocusToFirstItem();
+        });
+    }
+
+    ngOnDestroy(): void {}
+
+    private registerRootMenuContainer(): void {
+        if (this.isRootDropdown) {
+            const rootMenuContainer = this.hostVcdDropdown._clrDropdownMenuEl;
```
ryan-bradford, line 91: Instead of relying that `this.hostVcdDropdown._clrDropdownMenuEl` is set, would it be possible to pass the `rootMenuContainer` as an argument to this method and call it from the dropdown component when _clrDropdownMenuEl is set? Same for the place that you use `_dropdownTriggerEl`? Just relying that some internal property within this.hostVcdDropdown is set feels strange.

ryan-bradford (changes requested):
Just a few ideas and general comments. Some of them were just thoughts I had and not sure if they're totally possible, free to talk if you want

projects/components/src/dropdown/dropdown.component.ts
juanmendes: It's usually safer to create a function that returns an object. This creates a new copy every time it's called instead of everyone always sharing an object

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
ps37: This is just a trigger that opened this menu

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
ps37: I will add a comment to the menu trigger property

ryan-bradford (approved):

projects/components/src/dropdown/dropdown.component.ts
```diff
@@ -88,7 +109,39 @@ export class DropdownComponent<T extends DropdownItem<T>> {
...
+        this._clrDropdownMenu.items.reset([]);
+        this._clrDropdownMenu.items.notifyOnChanges();
+    }
+    private _clrDropdownMenu: ClrDropdownMenu;
+
+    @ViewChild(ClrDropdownTrigger, { read: ElementRef, static: true })
+    set dropdownTriggerEl(val: ElementRef<HTMLElement>) {
+        this._dropdownTriggerEl = val ? val.nativeElement : null;
+    }
+    _dropdownTriggerEl: HTMLElement;
```
juanmendes, line 131: private?

juanmendes (approved):

projects/components/src/dropdown/dropdown.component.ts
ps37: It is being accessed from the outside in dropdown-focus-handler.directive.ts

PR #225, opened 2020-10-22
Add more examples to show all the features of action menu

ps37:
The existing examples page does not do the justice of explaining all the features of the action menu component. In this change, more examples are added.

## Testing done:
actionMenuExamples


projects/components/src/assets/resources/en.properties
```diff
@@ -73,6 +73,7 @@ vcd.cc.numeric.filter.to=To:
 vcd.cc.datagrid.actions=Actions
 
 vcd.cc.action.menu.actions=All Actions
+vcd.cc.action.menu.other.actions=Other Actions
```
ryan-bradford, line 76: Because this fix is being merged into master and that is the branch that will be used in 10.2.1 (is this true still?), are there any i18n restrictions for patch releases? Is there anything extra we need to do via communicating this to Sarah?

ryan-bradford (approved):
One small i18n comment

projects/components/src/assets/resources/en.properties
juanmendes: We'll have a separate branch for VCD's 10.2.1, which we'll call a11y here

juanmendes:
Yumeng is creating a schematic that will put examples into a new folder. Please talk to him where these should be going and the naming folder conventions, I think it's `projects/examples/src/app/components`. I've asked @yumengwu  to add a comment with specific directions

yumengwu:
The current design for examples' structure is like `show-clipped-text` example. The path is: `projects/examples/src/app/components/show-clipped-text`.

For example, when adding two `action-menu`'s examples 'action menu one' and 'action menu two', there will be an `action-menu` folder under `projects/examples/src/app/components`.

`projects/examples/src/app/components/action-menu` folder contains one module, `action-menu-examples.module.ts`, which imports sub module and registers documentation entry with statement `Documentation.registerDocumentationEntry`, and two sub folders, `one` and `two`, which contain components' html, scss and ts files, and a module having `ActionMenuOne(Two)Example` component as exports and entry components.

juanmendes:
> The current design for examples' structure is like `show-clipped-text` example. The path is: `projects/examples/src/app/components/show-clipped-text`.
> 
> For example, when adding two `action-menu`'s examples 'action menu one' and 'action menu two', there will be an `action-menu` folder under `projects/examples/src/app/components`.
> 
> `projects/examples/src/app/components/action-menu` folder contains one module, `action-menu-examples.module.ts`, which imports sub module and registers documentation entry with statement `Documentation.registerDocumentationEntry`, and two sub folders, `one` and `two`, which contain components' html, scss and ts files, and a module having `ActionMenuOne(Two)Example` component as exports and entry components.

Just remember that `show-clipped-text-directive` is currently in the wrong folder.

PR #227, opened 2020-10-23
RFC: Rework the Widget Object

ryan-bradford:
# Summary

We set out with the goal to have one widget object for both Cypress E2E and Angular Unit tests. We initially thought this could be done through a system of promises and assertions that would force E2E tests to look like Angular tests. However, after much research, it seems as though it's not good design to `await` a Cypress Chainable. Additionally, we received some backlash from the team about the old design (more on that later). With all that in mind, we designed this new system of `WidgetObject`, where there is a shared `LocatorDriver` between the two types of tests that is responsible for locating DOM elements, and the tests will run the actions and assertions.

# PR Contents (bolded is important)

1. Deleting old widget-object pattern.
2. **Writing new widget-object pattern**
3. **Converting the Data Exporter Widget to use this new widget object**
4. **Updating the unit test and adding in a Cypress E2E test for the Data Exporter.**

-----------------
# Design

There are 6 important classes exposed to the client of this library as seen in [index.ts](projects/components/src/utils/test/locator/index.ts). These four classes are as follows:

1. `BaseWidgetObject<T>` - What your widget objects will extend 

This is the class that all widget objects will extend. It has a `LocatorDriver` which knows how to locate elements in either Cypress or Jasmine/Karma. The three location methods of this driver return another `LocatorDriver`, meaning these methods can (and should) be chained to make finding DOM elements easy. This locator driver has five important methods:
---- ***Chainable Locators***
a. get: Finds all child elements that matches the given `cssSelector`. This is equivalent to Cypress `find`.
b. getByText: Finds all child elements that match the given `cssSelector` and have text that contains the given `value`. This is equivalent to Cypress `contains`.
c. parents: Finds the closest parent that matches the given `cssSelector`. This is equivalent to Cypress `parents`.
---- ***Other***
d. findWidget: Finds another `BaseWidgetObject` within this locator. 
c. unwrap: Takes the result of the built query and returns the underlying type T. In the case of Cypress, this will be a `Chainable`, in the case of Angular, this will be a `TestElement` (more on this below). 

These 5 methods are what will be used to build all of your shared widget objects and have the capacity to query for any element in the DOM. 

2. `TestElement` -- What Angular widgets will return. 

Because any query built using the `LocatorDriver<T>` can return many elements, it was necessary to make some `DebugElement` wrapper that could store many elements, just as the `Cypress` Chainable can store many elements. The interface for this wrapper was modeled after the `TestElement` from the @anular/cdk, and the JQuery element, and the Cypress Chainable. There are a few methods in this interface to take actions like click, getText, clearValue, and other such things. 
***This interface is not complete yet***. Please suggest additions to the `TestElement`.

3/4: `WidgetObjectFinder` -- How you will find your widgets.

Once you have designed your `WidgetObject`, you need to use it in a test. To do this, you will use either the `CypressWidgetObjectFinder` or the `AngularWidgetObjectFinder` depending on your context. These classes know how to find your element given a correct `BaseWidgetObject<T>` and properly instantiate it's `LocatorDriver` and add the correct type information to make writing tests a breeze. 

5. `FindableWidget<T>` - What the widget locator excepts

This is an interface that enforces that a class extends `BaseWidgetObject<T>` and has a `static tagName: string` property. This means that the class is a widget that can be located using its tagName by the above finders.

6. `findCypressWidget` - A connivence locator function.

This is a function that can be used to locate a Cypress widget given some `FindableWidget<T>`. It is a simple connivence function that wraps the `CypressWidgetFinder`.

-----------------
# Implementing Tests
With this new pattern, writing tests will now have the following steps:

1. Define Widget Objects for any of the important, reusable elements in the Component you want to test. For example, if you are testing some component that contains a Datagrid, you should define widget objects for both your component and the `vcd-datagrid`. Both of these can then be reused by developers in either type of test when needed. 
2. Use the correct Widget Object Finder to find your widget object. 
3. Write tests in the pattern that your framework wants. This means in Cypress, all assertions should happen using the `should` method, and all methods on one found widget should be chained. ***However,*** do not use any querying methods outside of the widget object, if possible. For Angular Unit tests, all assertions should happen in-line and through the `expect` function. Additionally, use the connivance methods on the `TestElement` and ***do not*** access the debugElements stored within, if possible. 

Example tests are as follows:

 ```
// e2e test
it("enables ok button", () => {
    const loginWidgetObject = findCypressWidget(LoginWidgetObject);
    loginWidgetObject.getOkButton().should('be.disabled');
    loginWidgetObject.getUserName().type('username');
    loginWidgetObject.getPassword().type('a$h7r4y');
    loginWidgetObject.getOkButton().should('be.enabled');
})
// unit test
it("enables ok button", () => {
    const loginWidgetObject = new AngularWidgetObjectFinder(LoginComponent).find(LoginWidgetObject);
    expect(loginWidgetObject.getOkButton().enabled()).toBe(false);
    loginWidgetObject.getUserName().type('username');
    loginWidgetObject.getPassword().type('a$h7r4y');
    expect(loginWidgetObject.getOkButton().enabled()).toBe(false);
})
```

-----------------
# Why This?

We initially intended for both tests to look similar and use the same system of assertions and awaits. However, there were two reasons we switched away from this:
1. Cypress doesn't approve of using awaits, and it doesn't approve of conditional testing. This means that they do not want tests that run `if A then B else C`. This is because they want their tests to be deterministic and less-flaky. While we are unsure of their reasoning behind this, the `cy` object started throwing errors that were hard to get around when we used `awaits`. We might have been able to hack around this, but it would have been a hack. Instead, with this way of doing things, we are conforming to Cypress's standards.
2. We received feedback from some developers that they liked Cypress's system for finding and working with elements. Namely, the system of using Chainables to find and work with elements. Our old system of `WidgetObject` would have prevented developers from taking advantage of this. Instead, we found a way to allow developers to harness the power of Chainables in both Cypress and Angular!

-----------------
# Pros/Cons

**Pros**
1. Allows developers to use chainables to find elements in both Cypress and Karma/Jasmine.
2. Allows developers to use Cypress's system of assertions, and use Angular's system of assertions when appropriate.
3. One widget object for both types of tests! (this has a ton of benefits, that I can get into if we want...)

**Cons**
1. Developers can't make methods like `getHeaderText` or `clickLoginButton` in the widget object. This is because all the widget is responsible for is finding elements - nothing else. We feel as though this con isn't too bad because `widget.getLoginButton().click()` is just as generalized as `widget.clickLoginButton()`. 

The more extreme case of this is if you want to make a method like `loginUser(username, password)` that runs many steps on the component. Again, this is not possible because the widget can only *find* elements. It cannot perform actions. However, we decided that widgets should not contain business logic, and this logic should be represented in the tests. So in some ways, this con is a pro in that it enforces this claim to be true.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -41,13 +41,16 @@
     "@clr/ui": "3.1.3",
     "@ngx-formly/core": "5.6.1",
     "@stackblitz/sdk": "1.3.0",
+    "@types/jquery": "^3.5.3",
```
juanmendes, line 44: Please use `--save-exact`

projects/components/src/utils/test/locator/widget-locator.ts
```diff
@@ -0,0 +1,82 @@
...
+ *      getOkButton(): T {
+ *          return this.driver.getByText('button', 'Ok').unwrap();
+ *      }
+ *
+ *      getOkButtonContainer(): T {
+ *          return this.driver.getByText('button', 'Ok').parent('.button-container').unwrap();
+ *      }
+ * }
+ */
+export class BaseWidgetLocator<T> {
```
juanmendes, line 29: The base class is a widget object, not a locator, it uses a locator. 

I think the base object should just be `WidgetObject`

projects/components/src/utils/test/locator/widget-locator.ts
```diff
@@ -0,0 +1,82 @@
...
+export type CorrectReturnTypes<T, R> = {
+    [P in keyof T]: T[P] extends (...args: any[]) => infer CR
+        ? (...args: Parameters<T[P]>) => ConvertGenericOnly<CR, R>
+        : ConvertGenericOnly<T[P], R>;
+};
+
+/**
+ * A locator that is findable by tagName.
+ */
+export interface FindableLocator<T> extends Type<T> {
```
juanmendes, line 80: The locator name is being overused, the only thing that is a locator is the driver that accepts queries. This is trying to find a widget

projects/components/src/utils/test/locator/widget-locator.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Knows how to find elements of a given type T in the DOM.
+ */
+export interface LocatorDriver<T> {
+    /**
+     * Finds all child elements that match the given {@param cssSelector}.
+     */
+    get(cssSelector: string): LocatorDriver<T>;
+
+    /**
+     * Finds all child elements that match the given {@param cssSelector} and have text that contains the given {@param value}.
```
juanmendes, line 46: I haven't heard of embedding `@param` within the description. They should be separate lines. Please add `@param` for each of the functions

projects/components/src/utils/test/locator/cypress-widget-locator.ts
```diff
@@ -0,0 +1,96 @@
...
+import { FindableWidget, FindParams, isFindParamsObject } from '../widget-object';
+import { BaseWidgetLocator, CorrectReturnTypes, FindableLocator, LocatorDriver } from './widget-locator';
+
+type Chainable = Cypress.Chainable<JQuery<HTMLElement>>;
+
+// TODO: use alias to find widgets within other widgets
+/**
+ * Knows how to find Cypress chainables in the DOM.
+ */
+export class CypressLocatorDriver implements LocatorDriver<Chainable> {
```
juanmendes, line 16: There's no need for driver in the name, it's just a locator. Driver was when it was responsible for driving the test (clicking, typing)

projects/components/src/utils/test/locator/cypress-widget-locator.ts
```diff
@@ -0,0 +1,96 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { IdGenerator } from '../../id-generator/id-generator';
+import { FindableWidget, FindParams, isFindParamsObject } from '../widget-object';
+import { BaseWidgetLocator, CorrectReturnTypes, FindableLocator, LocatorDriver } from './widget-locator';
+
+type Chainable = Cypress.Chainable<JQuery<HTMLElement>>;
+
+// TODO: use alias to find widgets within other widgets
```
juanmendes, line 12: Is this comment still valid? Looks like we are using aliases

projects/components/src/utils/test/locator/widget-locator.ts
juanmendes: Talked as a team and the new suggestion is `WidgetElementLocator`. Please make sure documentation specifies that it should not try to bundle actions, such as click, type, into the class.

projects/components/src/utils/test/locator/cypress-widget-locator.ts
ryan-bradford: I think based on our conversation we decided to keep this name the same, but let me know if I interpreted that

projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
```diff
@@ -0,0 +1,40 @@
...
+export class CypressWidgetObjectFinder {
+    /**
+     * Finds a single widget object
+     *
+     * @param widgetConstructor the constructor of the widget to use
+     * @param ancestor the parent to begin the search from
+     * @param cssSelector the cssSelector to post-pend to the tagName for the search
+     *
+     */
+    public find<T extends TaggedClass & Type<BaseWidgetObject<Chainable>>>(
```
yumengwu, line 25: I am wondering if it is necessary to create new CypressWidgetObjectFinder object every time using find function like `new CypressWidgetObjectFinder`. Since this class only has this function, it seems like a static function or export function is enough to find widget object.

juanmendes (approved):

projects/components/src/utils/test/widget-object/widget-object.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Changes the class so that all functions/getters return R.
+ */
+export type CorrectReturnTypes<T, R> = {
+    [P in keyof T]: T[P] extends (...args: any[]) => infer CR
+        ? (...args: Parameters<T[P]>) => ConvertGenericOnly<CR, R>
+        : ConvertGenericOnly<T[P], R>;
+};
+
+/**
+ * A class with a static tagName.
```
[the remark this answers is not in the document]
ryan-bradford: Went with the latter stackoverflow idea.

projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
ryan-bradford: I tried to implement this, and while in Cypress it might make sense to have this just be a function, the AngularWidgetFinder exposes useful information in tests. Namely, it's what allows the user to run detectChanges and get access to the component if needed (like in the data-exporter tests). Furthermore, the Angular finder also creates the component in its constructor if given a component type, and this conceptually would be weird to move to the `find` method. 

So I guess we could make the Cypress finder a function and keep the Angular finder a class, but that feels unnecessary. For now, I'm going to keep the finder's unchanged as far as this comment goes, but let me know what you think.

projects/components/src/utils/test/widget-object/widget-object.ts
```diff
@@ -0,0 +1,106 @@
...
+import { Type } from '@angular/core';
+
+/**
+ * A class that is able to make queries to the DOM and be instantiated by a widget finder.
+ * Widget objects should be used only to find elements in the DOM. You should not attempt to take
+ * actions on these elements in a given widget object.
+ *
+ * All widget objects should extend this base class.
+ *
+ * @example
```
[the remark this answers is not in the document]
ryan-bradford: Example does display correctly: 
image

projects/components/src/utils/test/widget-object/angular-widget-object.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+        this.fixture.detectChanges();
+        this.clickTracker = new ClickTrackerWidgetObject(
+            new AngularLocatorDriver(
+                new TestElement([this.fixture.debugElement], this.fixture),
+                this.fixture.debugElement
+            )
+        );
+    });
+
+    afterEach(function (this: HasClickTracker): void {
```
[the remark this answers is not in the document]
ryan-bradford: Apparently not! Removed

projects/components/src/utils/test/widget-object/angular-widget-object.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (this.fixture) {
+            this.fixture.destroy();
+        }
+    });
+
+    describe('constructor', () => {
+        it('can be called when you create the instance directly', function (this: HasClickTracker): void {
```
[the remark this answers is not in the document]
ryan-bradford: Yea, I think you're right, this test does seem pretty trivial. Removing it

projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
```diff
@@ -0,0 +1,40 @@
...
+        ancestor?: string,
+        cssSelector?: string
+    ): CorrectReturnTypes<InstanceType<T>, Chainable> {
+        let tagName = widgetConstructor.tagName;
+        if (cssSelector) {
+            tagName += `${cssSelector}`;
+        }
+        const id = idGenerator.generate();
+        const search = ancestor ? cy.get(ancestor).find(tagName) : cy.get(tagName);
+        const promise = search.as(id);
```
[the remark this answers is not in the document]
ryan-bradford: Root is much better; changed

projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
yumengwu: Yes, since angular finder class is more complex, I think we can keep this for consistency

PR #228, opened 2020-10-26
[a11y] Add aria attributes to the Form Input component

georgiatanasov147:
- Added aria-required which depends on the 'showAsterisk' input.
The screen reader will announce if the field is required.
- Added aria-describedby for either errors or description in the input subtext.
- Added ids to the description and error elements using the IdGenerator.
- Fixed the 'for' attribute binding to the input label

Testing done:
- Tested with NVDA on the examples project
- Unit tests for each case

### DIFF
--- a/projects/components/src/form/base-form-control.ts
+++ b/projects/components/src/form/base-form-control.ts
@@ -18,7 +18,17 @@ export class BaseFormControl implements ControlValueAccessor, CanBeReadOnly {
     /**
      * Auto generated ID for the input field.
      */
-    id: string;
+    readonly id: string = idGenerator.generate();
+
+    /**
+     * Auto generated ID for the description field.
+     */
+    readonly descriptionId: string = `${this.id}-description`;
+
+    /**
+     * Auto generated ID for the error field.
+     */
+    readonly errorsId: string = `${this.id}-errors`;
 
     /**
      * Change callback.
@@ -72,7 +82,6 @@ export class BaseFormControl implements ControlValueAccessor, CanBeReadOnly {
     protected initialValue: number | string | boolean;
 
     constructor(ngControl: NgControl) {
-        this.id = idGenerator.generate();
         if (ngControl) {
             ngControl.valueAccessor = this;
             this.formControlNameDirective = ngControl;
--- a/projects/components/src/form/form-input/form-input.component.html
+++ b/projects/components/src/form/form-input/form-input.component.html
@@ -1,6 +1,6 @@
 <div class="form-group">
     <div class="clr-form-control">
-        <label *ngIf="label" class="clr-control-label" [attr.for]="id" [ngClass]="{ 'required-field': showAsterisk }">{{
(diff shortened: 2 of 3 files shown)


projects/components/src/form/form-input/form-input.component.html
```diff
@@ -17,18 +17,20 @@
                     [min]="min"
                     [max]="max"
                     [attr.maxlength]="maxlength"
+                    [attr.aria-required]="showAsterisk"
+                    [attr.aria-describedby]="showErrors ? errorsId : descriptionId"
```
irahov, line 21: descriptionId is valid only if there is a description

projects/components/src/form/form-input/form-input.component.html
```diff
@@ -17,18 +17,20 @@
                     [min]="min"
                     [max]="max"
                     [attr.maxlength]="maxlength"
+                    [attr.aria-required]="showAsterisk"
```
irahov, line 20: I would also add some unit tests

irahov (approved):
Please prepend with [a11y]

projects/components/src/form/form-input/form-input.component.html
georgiatanasov147: It appears we always have a value for `description`. In the `BaseFormControl` the `@Input` is initialized: 

```
/**
 * The description that appears below the control
 * \u00A0 is &nbsp; (non-breaking space)
 */
@Input() description = '\u00A0';
```

projects/components/src/form/form-input/form-input.component.html
georgiatanasov147: Just added unit tests.

projects/components/src/form/base-form-control.ts
```diff
@@ -73,6 +83,8 @@ export class BaseFormControl implements ControlValueAccessor, CanBeReadOnly {
 
     constructor(ngControl: NgControl) {
         this.id = idGenerator.generate();
+        this.descriptionId = idGenerator.generate();
```
juanmendes, line 86: We don't really new a new ID, it could be
```ts
this.descriptionId = `${this.id}-description`;
this.errorsId =`${this.id}-errors`;
```

Also consider moving them to the initializer, makes for less typing and simpler constructors. Adding readonly will also prevent it from being reassigned

```ts

    /**
     * Auto generated ID for the input field.	     
    */
    readonly id = idGenerator.generate();
  /**
     * Auto generated ID for the description field.
     */
   readonly  descriptionId = `${this.id}-description;

    /**
     * Auto generated ID for the error field.
     */
    readonly errorsId = `${this.id}-errors`;
```

juanmendes (approved):

projects/components/src/form/base-form-control.ts
georgiatanasov147: Done, thanks!

PR #240, opened 2020-10-30
Remove button config interface as we no longer support that API

ps37:
## Changes:
Button config is replaced with Action menu and actions and we no longer require that. It was not cleaned previously when the Action menu and actions API was introduced. this review cleans that and also fixes the Activity reporter example of data grid component which uses the button config API.

## Testing done:
replaceButtonConfig


projects/components/src/datagrid/interfaces/datagrid-column.interface.ts
```diff
@@ -44,132 +44,6 @@ export interface FunctionRenderer<T> {
...
-     */
-    globalButtons: GlobalButton<R>[];
-    /**
-     * The buttons whose displayability depends on the selected entity.
-     */
-    contextualButtonConfig?: ContextualButtonConfig<R>;
-    /**
-     * The way buttons should be shown when inactive.
-     */
-    inactiveDisplayMode?: InactiveButtonDisplayMode;
```
juanmendes, line 170: Shouldn't `InactiveButtonDisplayMode` also be deleted?

juanmendes (approved):

PR #248, opened 2020-11-10
Fixes the type clash between Jasmine and Chai in unit tests

ryan-bradford:
# Problem

When we loaded the Cypress types in our library via either `types: [..., "cypress"]` in the tsconfig or via `import * from 'cypress'` in a given file, it loads the `global namespace Cypress`. This also loads the global namespace for Chai (and Sinon and a few other libraries). This namespace was then loaded whenever a user imported anything from `@vcd/ui-components`. This meant that in unit tests Jasmine and Chai would have a clash on definitions. 

# Fix

Force users to supply the Cypress types externally. Meaning, `CypressWidgetFinder` now has one generic type T of what type the underlying locator will find (documentation more clear in the code). In the scope of our library, we don't have to load Cypress anymore because the typing is supplied externally. Then in the stepdef file, all the user has to do is `new CypressWidgetObjectFinder<**Cypress.Chainable**>().find(DataExporterWidgetObject);`. The only new part is what is in bold. The only side effect of this is that it no longer makes sense to have a function that wraps around the `CypressWidgetFinder` because the generics were just too confusing. 

# Other Options

1. Turn just the CypressWidgetFinder into its own project. This felt like a clunky solution and would require a lot of maintenance work. 
2. Add a sub-package (https://github.com/ng-packagr/ng-packagr/blob/master/docs/secondary-entrypoints.md). This might be better than the above idea because it is less hard to maintain, but for a few reasons I just couldn't get it to work. Most of this was due to how it scopes files when you use a sub-package. 
3. Set `skipLibCheck: true` to the `tsconfig.spec.ts` in any application that uses our library. Not the best because we are removing type safety.
4. Some other way that allows us to load a global namespace scoped to only one file. I just couldn't get this one to work. I tried for a **long** time too.

### DIFF
--- a/projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
+++ b/projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
@@ -12,6 +12,12 @@ const idGenerator = new IdGenerator('cy-id');
 
 /**
  * Knows how to find Cypress widget objects within the DOM.
+ *
+ * T is the type of data that this will output when you call `unwrap`.
+ * In almost all cases, this should be a `Cypress.Chainable`.
+ * We chose to provide this as a generic because for reasons seen in
+ * [this PR](https://github.com/vmware/vmware-cloud-director-ui-components/pull/248)
+ * we could not load the Cypress types in our library.
  */
 export class CypressWidgetObjectFinder<T> {
     /**
--- a/projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
+++ b/projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
@@ -26,7 +26,7 @@ export class CypressWidgetObjectFinder<T> {
      * @param widgetConstructor - The constructor of the widget to use
      * @param ancestor - The CSS query or alias of the parent to begin the search from.
      *                 this will be passed to `cy.get` and is a global search.
-     * @param cssSelector - The cssSelector to post-pend to the tagName for the search
+     * @param cssSelector - The cssSelector to append to the tagName for the search
      *
      */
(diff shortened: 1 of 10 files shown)


projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
```diff
@@ -35,23 +42,6 @@ export class CypressWidgetObjectFinder {
         const search = ancestor ? cy.get(ancestor).find(tagName) : cy.get(tagName);
         const root = search.as(id);
         const widget = new widgetConstructor(new CypressLocatorDriver(root, true, id));
-        return (widget as any) as CorrectReturnTypes<InstanceType<C>, Chainable>;
+        return (widget as any) as CorrectReturnTypes<InstanceType<C>, T>;
     }
 }
-
-/**
```
juanmendes, line 42: Instead of just deleting the utility, let's move it to somewhere that only unit tests can access and vcd_ui can just copy paste the approach

projects/components/src/utils/test/widget-object/cypress-widget-finder.ts
```diff
@@ -35,23 +42,6 @@ export class CypressWidgetObjectFinder {
...
- * Finds a single widget object
- *
- * @param widgetConstructor - The constructor of the widget to use
- * @param ancestor - The CSS query or alias of the parent to begin the search from.
- *                 this will be passed to `cy.get` and is a global search.
- * @param cssSelector - The cssSelector to post-pend to the tagName for the search
```
juanmendes, line 48: nit: post-pend => append. I saw it in a few places.

juanmendes (approved):

ps37 (approved):

PR #250, opened 2020-11-10
[A11y] Make form inputs responsive

juanmendes:
## Description

Followed Clarity's pattern of setting CSS grid classes. This allows
form inputs to be visible when zoom is set to 200% by moving labels and inputs to be on their own lines under small resolutions.

* The aside content is moved into the clr-input-container so that is always stays next to it
* number with unit input added some custom CSS to keep the select dropdown for the unit, and the hint signpost on the same line

## Testing Done
Check the following pages.
* [vcd-form-input](https://vmware.github.io/vmware-cloud-director-ui-components/formInputs/example)
* [vcd-form-select](https://vmware.github.io/vmware-cloud-director-ui-components/formSelect/documentation)
* [vcd-number-with-unit-input](https://vmware.github.io/vmware-cloud-director-ui-components/numberWithUnitFormInput/example/number-with-unit-form-input) and the [unit less example](https://vmware.github.io/vmware-cloud-director-ui-components/numberWithUnitFormInput/example/number-with-unit-form-input-unitless)
* [vcd-form-checkbox](https://vmware.github.io/vmware-cloud-director-ui-components/formCheckbox/example)

For all examples, the following behavior should apply
* Zooming the page to 200% with the screen at 1440×900 causes the labels and inputs to display on different lines 
* Description and validation errors display below the input, aligned with the label above it
* Signpost display next to the input (and text is word wrapped)

At regular zoom
* Inputs will behave mostly as they did before, except that making the page smaller will eventually display the label above the input field.

checkbox
form-select
form-input
number-with-uni

### DIFF
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+ - Make input elements responsive so that labels and the input are on separate lines on small screens. Alllows keyboard
+   users zoomed in at 200% to view the label and input instead of it being clipped.
+
 ## [1.0.0-a11y.2] - 2020-10-28
 
 ### Added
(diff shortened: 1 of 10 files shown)


irahov (approved):
Looks great

ttsvetanov-vmware:
Looks great. 
One small comment for form-select. When show a error we need also to show error icon some as vdc-form input
image

juanmendes:
> Looks great.
> One small comment for form-select. When show a error we need also to show error icon some as vdc-form input
> image

Agree, but it doesn't affect `a11y` and was not introduced by this change. Please file a separate issue

PR #257, opened 2020-11-23
[a11y] Add hints to checkbox component

juanmendes:
Also fixed some a bad previous commit:
* vcdResponsiveInput was added as a class name instead of an attribute
* Module for vcdResponsiveInput was incorrectly setup
* Make sure we only query for input labels and containers that are immediate children of clr-form-control

## Testing Done

- Verified the checkbox input hint's for both the regular checkbox and the toggle buttons.
- Installed this version in VCD's UI to make sure modules were working.

image
image

### DIFF
--- a/projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
+++ b/projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
@@ -15,7 +15,7 @@ import { AfterViewInit, Directive, ElementRef, HostBinding } from '@angular/core
     selector: '.clr-form-control[vcdResponsiveInput]',
 })
 export class ResponsiveInputDirective implements AfterViewInit {
-    @HostBinding('class.clr-row') private clrGridRow = true;
+    @HostBinding('class.clr-row') public clrGridRow = true;
     constructor(private el: ElementRef<HTMLElement>) {}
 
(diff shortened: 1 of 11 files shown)


irahov:
For future - better have 2 different commits to address 2 issues: hint & vcdResponsiveInput

projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
```diff
@@ -15,7 +15,7 @@ import { AfterViewInit, Directive, ElementRef, HostBinding } from '@angular/core
     selector: '.clr-form-control[vcdResponsiveInput]',
 })
 export class ResponsiveInputDirective implements AfterViewInit {
-    @HostBinding('class.clr-row') private clrGridRow = true;
+    @HostBinding('class.clr-row') public clrGridRow = true;
```
irahov, line 18: why is this needed?

irahov (approved):

projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
juanmendes: Adds `clr-row` to the host of the directive which is required for the responsive CSS behavior.

PR #258, opened 2020-11-24
[a11y] Display errors for invalid numbers

juanmendes:
## Problem 
This fixes an issue where the form control's value was null if the user input could not be parsed as a number. We did not display any validation error to the user explaining that the input is invalid.

## Solution
`form-input` uses an internal default validator that checks for HTML's [validity](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) for `badInput`

Note that I created getters for the internal form controls as a refactor. Therefore, the more relevant part of the review is https://github.com/vmware/vmware-cloud-director-ui-components/pull/258/commits/702b562ebb61be1d8b803a92f37b0f4a745d460c

`number-with-unit` also uses an internal default validator that checks the error state of its internal `form-input`

## Implementation notes
The internal `form-input` within `number-with-unit` does not display any errors but it does display the description.

### Form with unit example
* Displays "Invalid number" if user types `e`, `1e`, `-e` 
  * In Firefox, anything is allowed so typing abc also displays "Invalid number"
### Form with unit - unit-less example
* Type -1 into warn level
  * See message indicating that number must be 1-100
* Type `-1e`
  * See message saying "Invalid number"
* Toggle box that says "Validator shows *" 
  * Make sure message still says "Invalid number" 

### Form input
* Displays "Invalid number" if user types `e`, `1e`, `-e` 
  * In Firefox, anything is allowed so typing abc also displays "Invalid number"
* Make sure "Input is required" displays for empty required f
form-input-parsing
number-with-unit-parsing

### DIFF
--- a/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.html
+++ b/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.html
@@ -14,7 +14,7 @@
         [maxlength]="maxlength"
         [showAsterisk]="showAsterisk"
         [label]="label"
-        [description]="description"
+        [description]="formControl.valid && description"
     >
         <aside [ngSwitch]="unitOptions.length">
             <ng-container *ngSwitchCase="0">
--- a/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.html
+++ b/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.html
@@ -5,7 +5,7 @@
 <div *ngIf="!isReadOnly" class="input-number-with-unit">
     <vcd-form-input
         #limitedInput
-        [formControl]="formGroup.controls.limited"
+        [formControl]="numberInput"
         [placeholder]="placeholder"
         type="number"
         [size]="size"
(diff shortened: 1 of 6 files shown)


projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.ts
```diff
@@ -278,25 +285,29 @@ export class NumberWithUnitFormInputComponent
         if (!this.limitedInput) {
             return;
         }
+        // Hide errors in number input
         Object.defineProperty(this.limitedInput, 'showErrors', {
-            get: this.limitedInputShowErrors,
```
irahov, line 282: This was needed to show the errors styles for the input element

projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.html
```diff
@@ -14,7 +14,7 @@
         [maxlength]="maxlength"
         [showAsterisk]="showAsterisk"
         [label]="label"
-        [description]="description"
+        [description]="formControl.valid && description"
```
irahov, line 17: Don't change the description. The input automatically hides/shows the description/error.

juanmendes:
It turns out that we need a big refactor so that `number-with-unit` does not use other vcd form inputs internally. This will allow us to avoid hacks to hide labels, error icons and error messages and aligning things. Will create a different pull request

Issue #172, opened 2020-08-12
The vcd-error-banner component has margins


bryanvb:
The vcd-error-banner has 24px margins on left and right which the Clarity alerts don't have and the vcd-loading-banner doesn't have. These margins should be removed for consistency.

ryan-bradford:
Closed in https://github.com/vmware/vmware-cloud-director-ui-components/pull/259

PR #259, opened 2020-11-24
Remove margins from error banner

RiaMahoney:
Issue #172 The banner component margins were 24px and are now 0. 

## Testing Done
* Go to http://localhost:4200/errorBanner/example
* Verify that banner aligns with button above it

### Before
image

### After
 image

Signed off by Ria Mahoney mpatrice@vmware.com

### DIFF
--- a/projects/components/src/common/error/error-banner.component.scss
+++ b/projects/components/src/common/error/error-banner.component.scss
@@ -4,6 +4,6 @@
  */
 
 clr-alert ::ng-deep .alert {
-    margin-left: 24px;
-    margin-right: 24px;
+    margin-left: 0;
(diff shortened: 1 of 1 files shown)


projects/components/src/common/error/error-banner.component.scss
```diff
@@ -4,6 +4,6 @@
  */
 
 clr-alert ::ng-deep .alert {
-    margin-left: 24px;
-    margin-right: 24px;
+    margin-left: 0px;
```
juanmendes, line 7: 0 does not need a unit (here and below)

juanmendes (approved):

PR #260, opened 2020-11-24
Upgraded to Clarity 4 and Angular 10.

ryan-bradford:
# Description

Updates the common components library to use Angular 10 and Clarity 4. 

# Testing

1. Assured the unit tests pass.
2. Went to the datagrid examples
    1. Tested the selection examples and made sure selection was maintained across pages.
    2. Tested filtering and saw that the values were emitted correctly
    3. Observed that the custom renderer still rendered the components.
3. Tested that the error banner could be shown/hidden via the Activity reporter.
4. Ran the data-exporter E2E tests.
5. Tested quick search by searching for "hello"


projects/examples/e2e/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "../../../out-tsc/e2e",
     "module": "commonjs",
-    "target": "es5",
```
irahov, line 6: Why do we change the target?

package.json
```diff
@@ -56,9 +56,9 @@
     "properties": "1.2.1",
     "rbradford-compodoc": "1.1.11",
     "rxjs": "6.5.5",
-    "tsickle": "0.38.1",
-    "tslib": "1.10.0",
-    "zone.js": "0.10.2"
+    "tsickle": "0.39.1",
+    "tslib": "^2.0.0",
```
irahov, line 60: All the dependencies should be exact dependencies. 
There are others, including in other files that should be changed.

irahov (approved):
It will be better in the future if the commit related to styling is a different one, a separated one. Then the reviewer may concentrate only on a smaller change list

projects/examples/e2e/tsconfig.json
ryan-bradford: Mistake from early in the upgrade, good catch. Reverted

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -226,14 +226,18 @@ describe('DatagridComponent', () => {
...
 
             describe('@Input() datagridSelection', () => {
-                it('emits multiple rows when set to multi selection', function(this: HasFinderAndGrid): void {
+                it('emits multiple rows when set to multi selection', fakeAsync(function (
+                    this: HasFinderAndGrid
+                ): void {
                     this.finder.hostComponent.selectionType = GridSelectionType.Multi;
                     this.finder.detectChanges();
                     this.clrGridWidget.selectRow(0);
+                    tick();
```
juanmendes, line 235: Let's try `fixture.whenStable()`

juanmendes (approved):

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -286,22 +294,24 @@ describe('DatagridComponent', () => {
...
-                    it('in multi selection', async function(this: HasFinderAndGrid): Promise<void> {
+                    it('in multi selection', async function (this: HasFinderAndGrid): Promise<void> {
                         this.finder.hostComponent.selectionType = GridSelectionType.Multi;
                         this.finder.hostComponent.gridData = {
                             items: mockData,
                             totalItems: 2,
                         };
                         this.finder.detectChanges();
                         this.clrGridWidget.selectRow(0);
+                        await new Promise((resolve) => setTimeout(() => resolve()));
```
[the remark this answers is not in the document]
ryan-bradford: fixture.whenStable did not work in either case, but I did extract the impl of `timeout`.

PR #261, opened 2020-11-24
Rewrite `ClrDatagridWidget` with new Widget Object

yumengwu:
## Description
 - Rewrote `VcdDatagridWidget` and `ClrDatagridWidget` using new Widget Object pattern.
 - Rewrote datagrid component unit tests using new `ClrDatagridWidget`.
 - Rewrote helper method `createDatagridFilterTestHelper` and `createDatagridFilterTestHelperWithFinder` using new Widget Object.
 - Added convenience methods `sendKeyboardEvent`, `classes`, and `parent` to `TestElement`.
 - Removed `WithGridBoldRenderer`. The `getBoldText` is implemented in `ClrDatagridWidget.getCellStrong`.

## Test done:
Ran unit tests and assured that all the tests passed.

### DIFF
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 4. INTERNAL: Rewrote helper method about creating datagrid filter unit test with new Widget Object.
 ### Removed
 1. BREAKING: Removed `WithGridBoldRenderer`. The functionality of `WithGridBoldRenderer` has been implemented in `ClrDatagridWidget`.
+### Deprecated
+1. Marked `WidgetFinder` as deprecated.
 
 ## [1.0.0-dev.50] - 2020-12-2
 ### Fixed
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
-[internal]: Update `ng-live-docs` to 0.0.8 to enable `@vmw/ng-live-docs:add-example` schematic.
+1. INTERNAL: Update `ng-live-docs` to 0.0.8 to enable `@vmw/ng-live-docs:add-example` schematic.
 ### Changed
 1. BREAKING: Rewrote `VcdDatagridWidget` and `ClrDatagridWidget` with new Widget Object.
 2. Added convenience methods to the `TestElement`.
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [1.0.0-dev.51] - 2020-12-4
 ### Added
 1. INTERNAL: Update `ng-live-docs` to 0.0.8 to enable `@vmw/ng-live-docs:add-example` schematic.
 ### Changed
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 [internal]: Update `ng-live-docs` to 0.0.8 to enable `@vmw/ng-live-docs:add-example` schematic.
+### Changed
+1. BREAKING: Rewrote `VcdDatagridWidget` and `ClrDatagridWidget` with new Widget Object.
+2. Added convenience methods to the `TestElement`.
+3. INTERNAL: Rewrote `datagrid.component` unit tests using new widget object.
+4. INTERNAL: Rewrote helper method about creating datagrid filter unit test with new Widget Object.
+### Removed
+1. BREAKING: Removed `WithGridBoldRenderer`. The functionality of `WithGridBoldRenderer` has been implemented in `ClrDatagridWidget`.
 
 ## [1.0.0-dev.50] - 2020-12-2
 ### Fixed
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -53,8 +53,6 @@ interface HasFinderAndGrid {
     clrGridWidget: ClrDatagridWidgetObject<TestElement>;
     // The Widget Object for the VCD Datagrid
     vcdDatagrid: VcdDatagridWidgetObject<TestElement>;
-    // The instance of DatagridComponent
-    component: MockRecordDatagridComponent;
     hostComponent: HostWithDatagridComponent;
 }
 
@@ -76,10 +74,6 @@ describe('DatagridComponent', () => {
         // @ts-ignore
         this.vcdDatagrid = this.finder.find(VcdDatagridWidgetObject);
         this.clrGridWidget = this.vcdDatagrid.findClrDatagrid();
-        this.component = this.vcdDatagrid.unwrap().elements[0].componentInstance as DatagridComponent<
-            MockRecord,
-            RecordId
-        >;
         this.hostComponent = this.finder.hostComponent as HostWithDatagridComponent;
     });
 
@@ -124,7 +118,7 @@ describe('DatagridComponent', () => {
             it('gives proper css class for the grid', function (this: HasFinderAndGrid): void {
                 this.hostComponent.clrDatagridCssClass = 'some_class';
                 this.finder.detectChanges();
-                expect(this.clrGridWidget.unwrap().classes()).toContain('some_class');
+                expect(this.clrGridWidget.component.classes()).toContain('some_class');
             });
 
             it('sets no default CSS classnames for the rows', function (this: HasFinderAndGrid): void {
@@ -172,7 +166,7 @@ describe('DatagridComponent', () => {
                 it('clips text when disableCliptext is unset', function (this: HasFinderAndGrid): void {
                     this.hostComponent.columns = [{ displayName: 'Name', renderer: 'name' }];
                     this.finder.detectChanges();
-                    const res = this.clrGridWidget.getCell(0, 0).elements[0].injector.get(ShowClippedTextDirective);
+                    const res = this.clrGridWidget.getCell(0, 0).getInjector().get(ShowClippedTextDirective);
                     expect(res.disabled).toBeFalsy();
                 });
 
@@ -181,7 +175,7 @@ describe('DatagridComponent', () => {
                         { displayName: 'Name', renderer: 'name', cliptextConfig: { size: TooltipSize.md } },
                     ];
                     this.finder.detectChanges();
-                    const res = this.clrGridWidget.getCell(0, 0).elements[0].injector.get(ShowClippedTextDirective);
+                    const res = this.clrGridWidget.getCell(0, 0).getInjector().get(ShowClippedTextDirective);
                     expect(res.disabled).toBeFalsy();
                 });
 
@@ -190,7 +184,7 @@ describe('DatagridComponent', () => {
                         { displayName: 'Name', renderer: 'name', cliptextConfig: { disabled: true } },
                     ];
                     this.finder.detectChanges();
-                    const res = this.clrGridWidget.getCell(0, 0).elements[0].injector.get(ShowClippedTextDirective);
+                    const res = this.clrGridWidget.getCell(0, 0).getInjector().get(ShowClippedTextDirective);
                     expect(res.disabled).toBeTruthy();
                 });
             });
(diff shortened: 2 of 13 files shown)


projects/components/CHANGELOG.MD
```diff
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 [internal]: Update `ng-live-docs` to 0.0.8 to enable `@vmw/ng-live-docs:add-example` schematic.
+### Changed
+1. BREAKING: Rewrote `VcdDatagridWidget` and `ClrDatagridWidget` with new Widget Object.
```
ryan-bradford, line 12: Can you switch this to be the notation that Juan uses on line 10 with the brackets: []?

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -73,276 +72,293 @@ describe('DatagridComponent', () => {
             declarations: [HostWithDatagridComponent, DatagridDetailsComponent, DatagridDetailsPaneComponent],
         }).compileComponents();
 
-        this.finder = new WidgetFinder(HostWithDatagridComponent);
-        this.vcdDatagrid = this.finder.find(GridWithBoldRenderer);
-        this.clrGridWidget = this.vcdDatagrid.clrDatagrid;
-        this.component = this.finder.find(VcdDatagridWidgetObject).component as DatagridComponent<MockRecord, RecordId>;
+        this.finder = new AngularWidgetObjectFinder(HostWithDatagridComponent);
+        // @ts-ignore
+        this.vcdDatagrid = this.finder.find(VcdDatagridWidgetObject);
```
ryan-bradford, line 77: Was this line throwing an error without the `// @ts-ignore`? If so, we might want to work together to improve the types that come out of the AngularWidgetObjectFinder. I tried to use some magic types but they might not work perfectly.

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -73,276 +72,293 @@ describe('DatagridComponent', () => {
...
+        this.clrGridWidget = this.vcdDatagrid.findClrDatagrid();
+        this.component = this.vcdDatagrid.unwrap().elements[0].componentInstance as DatagridComponent<
```
ryan-bradford, line 79: Instead of doing this, we should rely on using this.finder.hostComponent to get access to all the Angular Components. (I think this should work, but let me know if it doesn't).

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -73,276 +72,293 @@ describe('DatagridComponent', () => {
...
                     };
                     this.finder.detectChanges();
                 });
 
                 it('clips text when disableCliptext is unset', function (this: HasFinderAndGrid): void {
-                    this.finder.hostComponent.columns = [{ displayName: 'Name', renderer: 'name' }];
+                    this.hostComponent.columns = [{ displayName: 'Name', renderer: 'name' }];
                     this.finder.detectChanges();
-                    expect(this.clrGridWidget.columnClippedTextDirective(0).disabled).toBeFalsy();
+                    const res = this.clrGridWidget.getCell(0, 0).elements[0].injector.get(ShowClippedTextDirective);
```
ryan-bradford, line 175: Instead of asserting this, can we assert that `text-overflow: ellipsis` is set on the style of the cell? When the directive is enabled it sets this property

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -478,123 +500,134 @@ describe('DatagridComponent', () => {
...
                         shouldShowPageSizeSelector: true,
                     };
                     this.finder.detectChanges();
-                    expect(this.clrGridWidget.getPaginationSizeSelectorText()).toEqual('Total Items52050100');
+                    expect(this.clrGridWidget.getPaginationSizeSelector().text()).toEqual('Total Items52050100');
                 });
             });
 
             describe('@Input() height', () => {
+                function getGridHeight(element: TestElement): string {
```
ryan-bradford, line 546: Good use of a helper function! Should we add some function to the test Element like `getStylePropertyValue(key: string)` to implement this instead?

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -478,123 +500,134 @@ describe('DatagridComponent', () => {
...
-                expect(this.clrGridWidget.placeholder).toEqual('Placeholder');
+                expect(this.clrGridWidget.unwrap().elements[0].componentInstance.loading).toBeFalsy();
+                expect(this.clrGridWidget.getPlaceHolder().text()).toEqual('Placeholder');
             });
 
             it('does not show the placeholder if the grid has data', function (this: HasFinderAndGrid): void {
                 this.finder.detectChanges();
-                expect(this.clrGridWidget.loading).toBeTruthy();
-                this.finder.hostComponent.gridData = {
+                expect(this.clrGridWidget.unwrap().elements[0].componentInstance.loading).toBeTruthy();
```
ryan-bradford, line 596: Is there anything else we can assert about the HTML to implement this instead of querying to the component?

ryan-bradford (changes requested):
This looks really great. Just a few small notes, and one overall comment. Pretty much, I think we should avoid using `componentInstance` as much as possible in our tests. Pretty much the only time when we should use the `componentInstance` is when we want to assert something about the component that isn't reflected in the HTML. So I think there were a few places where you used `componentInstance` that could be converted to querying the HTML, As with anything, totally let me know if you disagree.

projects/components/src/datagrid/datagrid.component.spec.ts
yumengwu: Without the `// @ts-ignore`, there is an error:
`TS2322: Type 'CorrectReturnTypes<VcdDatagridWidgetObject<unknown>, TestElement>' is not assignable to type 'VcdDatagridWidgetObject<TestElement>'.   The types returned by 'findClrDatagrid(...)' are incompatible between these types.     Type 'CorrectReturnTypes<ClrDatagridWidgetObject<unknown>, TestElement>' is missing the following properties from type 'ClrDatagridWidgetObject<TestElement>': _getRow, _getCell`

projects/components/CHANGELOG.MD
juanmendes: I don't remember doing line 10. We lose the list if we do that.

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -3,343 +3,256 @@
...
     /**
-     * Returns the CSS class names of the given Clarity datarow.
+     * Retrieves the cell
+     * @param row 0-based index of row
+     * @param col 0-based index of column
      */
-    getRowsCssClass(index: number): string[] {
-        return Object.keys(this.rows[index].classes);
+    getCell(row: number, col: number): T {
+        return this._getCell(row, col).unwrap();
```
juanmendes, line 85: Feels weird that we have to call `unwrap()` from so many places

projects/components/src/datagrid/datagrid.component.spec.ts
juanmendes, line 79: This line is used in 11 places `unwrap().elements[0].componentInstance` and looks hackish. I think we need to figure out a way to access the component instance more easily from the tests. 

We probably have to chat offline about this

projects/components/src/datagrid/datagrid.component.spec.ts
juanmendes: Setting text overflow is the job of the directive, we just need to make sure the directive is disabled

juanmendes: Having said this, `his.clrGridWidget.getCell(0, 0).elements[0].injector.get(ShowClippedTextDirective);` looks very hackish, let's talk about how to avoid this.

projects/components/src/utils/test/widget-object/angular-widget-object.ts
```diff
@@ -162,6 +162,37 @@ export class TestElement implements Iterable<TestElement> {
...
+     * Returns classes of first element as a string array
+     */
+    classes(): string[] {
+        return Object.keys(this.elements[0].classes);
+    }
+
+    /**
+     * Returns the parent of first element
+     */
+    parent(): TestElement {
```
juanmendes, line 192: This seems catered to a very specific example and we are using it for reaching outside of the widget object. Let's talk about how we can improve this

projects/components/src/utils/test/datagrid/datagrid.wo.ts
```diff
@@ -235,24 +236,7 @@ export class ClrDatagridWidgetObject<T> extends BaseWidgetObject<T> {
...
-     */
-    getRadioWrapper(): T {
-        return this.locatorDriver.get(Css.RADIO_WRAPPER).unwrap();
-    }
-
-    /**
-     * Unwrap the clr-datagrid
-     */
-    unwrap(): T {
+    get component(): T {
```
ryan-bradford, line 239: These methods make it seem like we're getting the component instance, can we rename these to refer to the name of the component? So this one would be `get clrDatagrid`?

projects/components/src/utils/test/datagrid/vcd-datagrid.wo.ts
```diff
@@ -28,7 +28,7 @@ export class VcdDatagridWidgetObject<T> extends BaseWidgetObject<T> {
     /**
      * Unwraps the `vcd-datagrid`
      */
-    unwrap(): T {
+    get component(): T {
```
ryan-bradford, line 31: same comment as above

projects/components/src/utils/test/datagrid/filter-utils.ts
```diff
@@ -75,9 +75,9 @@ export function createDatagridFilterTestHelper<V, C>(
     const grid = finder.find(ClrDatagridWidgetObject);
 
     wf.hostComponent.setFilter(filterType, wf, config || ({} as C));
-    grid.unwrap().fixture.detectChanges();
+    grid.component.fixture.detectChanges();
```
ryan-bradford, line 78: I think we still want to rework the filter tests to not rely on using component instances and to be built into the datagrid, but it might be a little much for this review. Can you add a ticket to our backlog titled something like "Rework datagrid filter tests to use WidgetObject rather than componentInstance"?

ryan-bradford (approved):

projects/components/src/utils/test/datagrid/filter-utils.ts
```diff
@@ -62,11 +69,15 @@ export function createDatagridFilterTestHelper<V, C>(
     configureTestingModule();
 
     // Add the filter to grid column
-    const finder = new WidgetFinder(FilterTestHostComponent);
+    const wf = new WidgetFinder(FilterTestHostComponent);
+
+    const finder = new AngularWidgetObjectFinder(FilterTestHostComponent);
```
juanmendes, line 74: I don't think we should use both `WidgetFinder` and `AngularWidgetFinder`, can't we just use `AngularWidgetFinder`?

projects/components/src/utils/test/datagrid/filter-utils.ts
```diff
@@ -16,8 +17,14 @@ import {
 import { MockRecord } from '../../../datagrid/mock-data';
 import { IdGenerator } from '../../id-generator/id-generator';
 import { WidgetFinder } from '../widget-object';
+import { AngularWidgetObjectFinder } from '../widget-object/angular-widget-finder';
+import { TestElement } from '../widget-object/angular-widget-object';
 import { ClrDatagridWidgetObject } from './datagrid.wo';
 
+function getFilter<V, C>(element: TestElement, filterType: Type<DatagridFilter<V, C>>): DatagridFilter<V, C> {
+    return element.elements[0].parent.parent.parent.query(By.directive(filterType)).componentInstance;
```
juanmendes, line 25: `parent.parent.parent` is very brittle, can't we use `parents()`?

projects/components/src/utils/test/widget-object/angular-widget-object.ts
```diff
@@ -208,6 +208,32 @@ export class TestElement implements Iterable<TestElement> {
         return this.elements[0].injector;
     }
 
+    /**
+     * Private helper function to find the first parent element that matches css selector
```
juanmendes, line 212: Please remove the words private helper function, they are already all implied/expressed in code.

```
Finds the first parent element that matches the CSS selector
```

projects/components/src/utils/test/widget-object/angular-widget-object.ts
```diff
@@ -208,6 +208,32 @@ export class TestElement implements Iterable<TestElement> {
...
+     */
+    private findParents(element: DebugElement, cssSelector: string): DebugElement {
+        if (!element) {
+            return undefined;
```
juanmendes, line 216: I'd rather return `null`. `undefined` means something that has been created but hasn't been assigned a value. In JS the two basic examples are when you do `let c;` and `a = {}; a.b`

`null` means something that has been assigned an empty value.

This way of searching is really expensive and wasteful because you're querying at each level of the DOM

The easier way to implement what you are doing is using [Element#matches](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches). In pseudo code:

```ts
if (!debugEl) {
   return null;
}
return debugEl.nativeElement.matches(cssSelector) 
    ? debugEl
    : this.findParent(this.debugEl.parent, cssSelector);
```

juanmendes (approved):

PR #262, opened 2020-11-24
Adds error icon to vcd-form-select

RiaMahoney:
## Testing Done

* Go to http://localhost:4200/formSelect/example
* Verify that error icon does not appear when user arrives on the page
* Verify that error icon does not appear when user selects an option from the drop-down
* Verify that error icon DOES appear when user fails to select an option from the drop-down

## Unit Tests Added

* Test to verify no icon shows when page loads
* Test to verify icon shows when user selects invalid option
* Test to verify no icon shows when user selects valid option

### Before
Screen Shot 2020-11-24 at 1 19 35 PM

### After - with error
Screen Shot 2020-11-24 at 4 47 43 PM

### After - without error
image

Signed-off-by: Ria Mahoney <mpatrice@vmware.com>

### DIFF
--- a/projects/components/src/form/form-select/form-select.component.html
+++ b/projects/components/src/form/form-select/form-select.component.html
@@ -13,6 +13,7 @@
                         {{ option.isTranslatable ? (option.display | translate) : option.display }}
                     </option>
                 </select>
+                <clr-icon *ngIf="showErrors" class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
             </div>
 
             <ng-content select="aside"></ng-content>
--- a/projects/components/src/form/form-select/form-select.component.spec.ts
+++ b/projects/components/src/form/form-select/form-select.component.spec.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { Component, ViewChild } from '@angular/core';
+import { Component, DebugElement, Type, ViewChild } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 import { SelectOption } from '../../common/interfaces/select-option';
 import { WidgetFinder, WidgetObject } from '../../utils/test/widget-object';
@@ -15,6 +15,15 @@ export class VcdFormSelectWidgetObject extends WidgetObject<FormSelectComponent>
 
     private get selectElement(): HTMLSelectElement {
         return this.findElement('select').nativeElement;
+    }
+
+    get clrIcon(): HTMLElement {
+        const clrIconDebugEl = this.findElement('clr-icon');
+        return clrIconDebugEl ? clrIconDebugEl.nativeElement : undefined;
+    }
+
+    findElement(selector: string | Type<unknown>, parent: DebugElement = this.root): DebugElement {
+        return super.findElement(selector, parent);
     }
 
     get value(): string {
@@ -81,6 +90,20 @@ describe('FormSelectComponent', () => {
             expect(selectInput.value).toEqual(hostComponent.options[1].value as string);
             hostComponent.formGroup.get('selectInput').setValue(hostComponent.options[2].value);
             expect(selectInput.value).toEqual(hostComponent.options[2].value as string);
+        });
+    });
+
+    describe('validation', () => {
+        it('should not show error icon before user selects value', () => {
+            expect(selectInput.clrIcon).toBeUndefined();
+        });
+        it('shows error icon when a invalid value is selected', () => {
+            selectInput.select(0);
+            expect(selectInput.clrIcon).toBeDefined();
+        });
+        it('does not show error icon when a valid value is selected', () => {
+            selectInput.select(1);
+            expect(selectInput.clrIcon).toBeUndefined();
         });
     });
 });
--- a/projects/components/src/form/form-select/form-select.component.spec.ts
+++ b/projects/components/src/form/form-select/form-select.component.spec.ts
@@ -19,7 +19,7 @@ export class VcdFormSelectWidgetObject extends WidgetObject<FormSelectComponent>
 
     get clrIcon(): string {
         const clrIconDebugEl = this.findElement('clr-icon');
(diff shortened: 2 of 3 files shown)


projects/components/src/common/error/error-banner.component.scss
```diff
@@ -4,6 +4,6 @@
  */
 
 clr-alert ::ng-deep .alert {
-    margin-left: 24px;
-    margin-right: 24px;
+    margin-left: 0;
```
ryan-bradford, line 7: Can you rebase off the most recent origin/master? This shouldn't show up in this diff because you did it in another commit.

projects/components/src/form/form-select/form-select.component.html
```diff
@@ -13,6 +13,7 @@
                         {{ option.isTranslatable ? (option.display | translate) : option.display }}
                     </option>
                 </select>
+                <clr-icon *ngIf="showErrors" class="clr-validate-icon" shape="exclamation-circle"></clr-icon>
```
ryan-bradford, line 16: Did you load this into VCD UI and test it out on any forms in there? It would be nice to get a screenshot of what this looks like in VCD UI

ryan-bradford (approved):

projects/components/src/common/error/error-banner.component.scss
RiaMahoney: Thanks. I rebased off master to take care of this.

projects/components/src/form/form-select/form-select.component.html
juanmendes: @ryan-bradford She will probable have to be guided on how to do that

juanmendes:
We should add a unit test for this. Please speak to @ps37 so he can help you create a unit test

projects/components/src/form/form-select/form-select.component.html
ryan-bradford: Done and done, we just had a good call

projects/components/src/form/form-select/form-select.component.spec.ts
```diff
@@ -83,6 +92,20 @@ describe('FormSelectComponent', () => {
             expect(selectInput.value).toEqual(hostComponent.options[2].value as string);
         });
     });
+
+    describe('validation', () => {
+        it('should not show error icon before user selects value', () => {
```
juanmendes, line 97: Please use `does not show error` instead of `should not show error icon`. We like to avoid repetitive text

projects/components/src/form/form-select/form-select.component.spec.ts
```diff
@@ -17,6 +17,15 @@ export class VcdFormSelectWidgetObject extends WidgetObject<FormSelectComponent>
         return this.findElement('select').nativeElement;
     }
 
+    get clrIcon(): HTMLElement {
```
juanmendes, line 20: We avoid returning an HTML element from a public method because that encourages querying outside the widget object wrapper. You probably want to return its [shape attribute](https://clarity.design/icons) as a string and you want to test that it has the correct icon

projects/components/src/form/form-select/form-select.component.spec.ts
```diff
@@ -17,6 +17,15 @@ export class VcdFormSelectWidgetObject extends WidgetObject<FormSelectComponent>
...
+        const clrIconDebugEl = this.findElement('clr-icon');
+        return clrIconDebugEl ? clrIconDebugEl.nativeElement : undefined;
+    }
+
+    findElement(selector: string | Type<unknown>, parent: DebugElement = this.root): DebugElement {
```
juanmendes, line 25: It's unclear why we need this method when it just passes arguments to its super

RiaMahoney: removed this method

projects/components/src/form/form-select/form-select.component.spec.ts
RiaMahoney: Changed to return shape attribute.

projects/components/src/form/form-select/form-select.component.spec.ts
```diff
@@ -17,6 +17,11 @@ export class VcdFormSelectWidgetObject extends WidgetObject<FormSelectComponent>
         return this.findElement('select').nativeElement;
     }
 
+    get clrIcon(): string {
```
juanmendes, line 20: Please rename to `clrIconShape`. Add some documentation explaining that it returns an empty string  if there's no icon

projects/components/src/form/form-select/form-select.component.spec.ts
```diff
@@ -17,6 +17,11 @@ export class VcdFormSelectWidgetObject extends WidgetObject<FormSelectComponent>
...
+        const clrIconDebugEl = this.findElement('clr-icon');
+        return clrIconDebugEl ? clrIconDebugEl.nativeElement.getAttribute('shape') : undefined;
```
juanmendes, line 22: This can be simplified to

```ts
return clrIconDebugEl?.nativeElement.getAttribute('shape') || '';
```

We usually prefer to return `null` instead of `undefined`. In this case an empty string makes sense.

projects/examples/src/app/home/home.component.spec.ts
```diff
@@ -7,7 +7,7 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { HomeComponent } from './home.component';
 
-describe('HomeComponent', () => {
+xdescribe('HomeComponent', () => {
```
juanmendes: Why is this being commented out?

projects/components/src/form/form-select/form-select.component.spec.ts
RiaMahoney: Simplified the return statement.

projects/components/src/form/form-select/form-select.component.spec.ts
RiaMahoney: done

PR #264, opened 2020-11-30
[master branch]Fix accessibility issues on Action menu component

ps37:
## Why this change:
Fix the following a11y issues in the action menu component:
- Navigating using up & down arrows on the action menu pop-up is making the underlying page scroll. The underlying page should not scroll.
- Pressing the Space/Enter keys on nested menus is not closing the action menu. Pressing those keys from anywhere on the menu should activate the button and then close the menu.
- Pressing the Esc key on the menu should just close the menu.
- Navigation using arrow keys on a menu with separators is not working.

## Testing done:
Tested the action menu in VM and Vapp action menus:
actionMenuA11yIssuesFix

In the examples app:
actionMenuA11yIssuesFix-1

### DIFF
--- a/projects/components/src/dropdown/dropdown-focus-handler.directive.ts
+++ b/projects/components/src/dropdown/dropdown-focus-handler.directive.ts
@@ -4,13 +4,23 @@
  */
 
 import { AfterViewInit, Directive, Host, OnDestroy, Optional, Renderer2, SkipSelf } from '@angular/core';
-import { ClrPopoverToggleService } from '@clr/angular';
-import { Subscription } from 'rxjs';
 import { SubscriptionTracker } from '../common/subscription';
 import { DropdownFocusHandlerService } from './dropdown-focus-handler.service';
 import { DropdownComponent } from './dropdown.component';
 
-const BUTTON_NODE_NAME = 'BUTTON';
+/**
+ * Html node names of different menu items other than separators, that can be part of a {@link DropdownComponent}s html
+ */
+enum DropdownMenuItemType {
+    /**
+     * Menu item that results in executing of an action when clicked
+     */
+    BUTTON = 'BUTTON',
+    /**
+     * Activating this would open a nested menu
+     */
+    NESTED_DROPDOWN_TRIGGER = 'VCD-DROPDOWN',
+}
 
 /**
  * Arrow keys directions
@@ -40,7 +50,7 @@ export interface MenuItem {
     /**
      * Call back to close the menu for which this menu item can be a trigger. called from {@link DropdownFocusHandlerService}
      */
-    closeMenu?: (event: Event) => void;
+    closeMenu?: () => void;
 }
 
 /**
@@ -59,10 +69,10 @@ export interface MenuItem {
 @Directive({
     selector: 'vcd-dropdown[vcdDropdownFocusHandler]',
 })
-export class DropdownFocusHandlerDirective<T> implements AfterViewInit, OnDestroy {
+export class DropdownFocusHandlerDirective implements AfterViewInit, OnDestroy {
     constructor(
         @Optional() @SkipSelf() private parentVcdDropdown: DropdownComponent,
-        @Optional() @SkipSelf() private parentFocusHandler: DropdownFocusHandlerDirective<T>,
+        @Optional() @SkipSelf() private parentFocusHandler: DropdownFocusHandlerDirective,
         @Host() private hostVcdDropdown: DropdownComponent,
         private focusHandlerService: DropdownFocusHandlerService,
         private renderer: Renderer2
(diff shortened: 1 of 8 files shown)


projects/components/src/dropdown/dropdown-focus-handler.service.ts
```diff
@@ -23,41 +23,47 @@ export class DropdownFocusHandlerService implements OnDestroy {
...
     }
 
     /**
-     * Calls the HTML focus method on the HTML element passed
+     * Calls the HTML focus method on the HTML element passed and removed focus on currently focused element.
+     * Returns true if the focus is moved and false otherwise.
      */
-    moveFocusTo(item: MenuItem): void {
+    moveFocusTo(item: MenuItem): boolean {
+        let moved = false;
```
ryan-bradford, line 52: Instead of having the moved variable, cant we just return a specific boolean from the places where we `return moved`. So the first two would be `return false` and the last one would be `return true`?

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -159,10 +169,18 @@ export class DropdownFocusHandlerDirective<T> implements AfterViewInit, OnDestro
...
-            left: this.menuTrigger,
-        }));
+        this.menuItems = menuChildren
+            .filter(
+                (child) =>
+                    child.nodeName === DropdownMenuItemType.BUTTON ||
+                    child.nodeName === DropdownMenuItemType.NESTED_DROPDOWN_TRIGGER
+            )
+            .map((child) => ({
+                element: (child.matches('button')
```
ryan-bradford, line 179: Should we make the part after `element:` a separate function? Call it something like `getDropdownButton(child)`?

projects/components/src/dropdown/dropdown-focus-handler.service.ts
```diff
@@ -66,28 +72,43 @@ export class DropdownFocusHandlerService implements OnDestroy {
...
+     * Whenever escape is pressed on dropdown item, the menu has to be closed and the focus has to be moved to parent menu if any... Just
+     * like when left arrow key is pressed
+     */
+    escapePressed(): void {
+        if (!this.currentFocusedItem.left) {
+            return;
+        }
+        // Close the nested menu before moving focus to left side
+        this.currentFocusedItem.left.closeMenu();
+        this.moveFocus(Direction.LEFT);
```
ryan-bradford, line 112: Will `this.moveFocus(Direction.LEFT);` also call `this.currentFocusedItem.left.closeMenu();`?

projects/components/src/dropdown/dropdown-focus-handler.service.ts
```diff
@@ -66,28 +72,43 @@ export class DropdownFocusHandlerService implements OnDestroy {
...
             })
         );
     }
+
+    /**
+     * Whenever escape is pressed on dropdown item, the menu has to be closed and the focus has to be moved to parent menu if any... Just
+     * like when left arrow key is pressed
+     */
+    escapePressed(): void {
+        if (!this.currentFocusedItem.left) {
```
ryan-bradford, line 107: If it's the parent element (in the case of this if statement), do we want to close the dropdown menu entirely?

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
```diff
@@ -4,13 +4,23 @@
...
+ */
+enum DropdownMenuItemType {
+    /**
+     * Menu item that results in executing of an action when clicked
+     */
+    BUTTON = 'BUTTON',
+    /**
+     * Activating this would open a nested menu
+     */
+    NESTED_DROPDOWN_TRIGGER = 'VCD-DROPDOWN',
```
ryan-bradford, line 22: Why is the string value all caps?

ryan-bradford (changes requested):
Just a few comments. One more thing is should we add some unit tests to test for the four bugs solved? Testing the scrolling one might be difficult

projects/components/src/dropdown/dropdown.component.ts
```diff
@@ -307,10 +311,28 @@ export class DropdownComponent implements AfterViewInit {
      * @param event Space or Enter key press event
      */
     onDropdownItemActivated(event: Event): void {
-        if (!event) {
+        event.stopPropagation();
+        event.target.dispatchEvent(new Event('click'));
```
irahov, line 315: a little bit of comment why we need to dispatch a click event rather than calling  onItemClicked(item)

irahov (approved):

projects/components/src/dropdown/dropdown-focus-handler.service.ts
ps37: No, movefocus method does not call the close menu

projects/components/src/dropdown/dropdown-focus-handler.service.ts
ps37: I assume you are asking what happens if the user presses the left arrow key on the first level of the dropdown. According to Clarity documentation, it has to close that level of dropdown which results in the closing of the entire dropdown

ps37:
> Just a few comments. One more thing is should we add some unit tests to test for the four bugs solved? Testing the scrolling one might be difficult

Added a Jira task to work on this later as I am trying to fix issues for the 10.2.1 release. https://jira.eng.vmware.com/browse/VDUCC-432

projects/components/src/dropdown/dropdown-focus-handler.directive.ts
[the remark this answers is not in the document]
juanmendes: `child.matches` is simpler

ryan-bradford (approved):

Issue #233, opened 2020-10-28
Setting datagridSelection to [] then getting datagridSelection results in [undefined] (in single selection grid).


bryanvb:
https://github.com/vmware/vmware-cloud-director-ui-components/blob/62db531d889c9af74a57a741c4b225c73b83612f/projects/components/src/datagrid/datagrid.component.ts#L451

selection[0] of [] is undefined so this.datagrid.selection.currentSingle = selection[0]; sets this.datagrid.selection.currentSingle to undefined.

Then:
https://github.com/vmware/vmware-cloud-director-ui-components/blob/62db531d889c9af74a57a741c4b225c73b83612f/projects/components/src/datagrid/datagrid.component.ts#L432

`
if (this.datagrid.selection.currentSingle) {
    return [this.datagrid.selection.currentSingle];
}
`

Will return [undefined].

bryanvb:
I was asked to add details about the impact of this bug so here is a summary. If a developer has a subscription to datagridSelectionChange and sets datagridSelection to [] then the subscription callback will be called with [undefined]. Since this is a pretty common use case I think it's high priority.

PR #270, opened 2020-12-04
Using datagridSelection getter to emit datagrid selection changes

RiaMahoney:
## Description
*  The problem appeared to be that this.datagrid.selection.currentSingle was set to undefined instead of an empty array when the user clears their selection.
* The template was getting datagridSelection from Clarity $event but it should have come from the getter.
* [$event] is an array of undefined. 
* Now if this.datagrid.selection.currentSingle is undefined, the logic in the getter handles for that and returns the empty array.

## Testing Done
* Added unit test to verify that when the selection is cleared an empty array is emitted.
* Added unit test to verify that when the selection is cleared the getter for the current selection returns an empty array.

Fixes: #233 

Signed-off-by: Ria Mahoney <mpatrice@vmware.com>

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -8,7 +8,6 @@ import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { LoadingListener } from '@clr/angular';
 import { MockTranslationService, TranslationService } from '@vcd/i18n';
-import { Mock } from 'protractor/built/driverProviders';
 import { Observable } from 'rxjs';
 import { first } from 'rxjs/operators';
 import { ActivityPromiseResolver } from '../common/activity-reporter/activity-promise-resolver';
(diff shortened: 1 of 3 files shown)


projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -446,9 +446,8 @@ export class DatagridComponent<R extends B, B = any> implements OnInit, AfterVie
             this.initialSelection = selection;
             return;
         }
-
         if (this._selectionType === GridSelectionType.Single) {
-            if (selection) {
+            if (selection && selection.length > 0) {
```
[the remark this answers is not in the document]
RiaMahoney: This change has been reverted.

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -243,6 +242,13 @@ describe('DatagridComponent', () => {
                     this.clrGridWidget.selectRow(1);
                     expect(this.finder.hostComponent.datagridSelection).toEqual([mockData[1]]);
                 });
+
+                it('reads an empty array and returns an empty array', function (this: HasFinderAndGrid): void {
+                    this.finder.hostComponent.selectionType = GridSelectionType.Single;
+                    this.finder.detectChanges();
+                    this.finder.hostComponent.datagridSelection = [];
+                    expect(this.finder.hostComponent.datagridSelection).toEqual([]);
```
ryan-bradford, line 250: Unless you run change detection between line 249 and 250, this test isn't asserting anything interesting. If you run change detection between these lines, it should allow your setter on line 249 to propagate to the datagrid component and for it to update the value accordingly. (reach out to me if you want any help with how change detection works, not that I'm an expert at all hahahaha)

ryan-bradford (approved):

projects/components/src/datagrid/datagrid.component.spec.ts
RiaMahoney: added this.finder.detectChanges(), ran tests again, all pass

PR #272, opened 2020-12-08
Update responsive input to increase label's width when container's width is small

yumengwu:
## Why this change is made
The reason for this issue is, in a responsive input, default clr grid classes for label and container are `clr-col-md-2` and `clr-col-md-10`, which are fine for most places. But when width for the host element row is small, there is no enough space for label.
This PR updates `ResponsiveInput` to add grid classes based on host element's width.

When host's width is less than 768px:
Screen Shot 2020-12-09 at 10 45 26 AM

When host's width is bigger than 768px:
Screen Shot 2020-12-09 at 10 46 16 AM

### DIFF
--- a/projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
+++ b/projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
@@ -3,29 +3,45 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { AfterViewInit, Directive, ElementRef, HostBinding } from '@angular/core';
+import { AfterViewChecked, Directive, ElementRef, HostBinding } from '@angular/core';
+
+declare type ColSize = '2' | '4' | '8' | '10';
 
 /**
- * Adds Clarity grid classes to form controls so that labels and inputs are on separate lines and to control the label width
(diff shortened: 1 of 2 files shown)


juanmendes:
Approved but please wait until https://github.com/vmware/vmware-cloud-director-ui-components/pull/273 is merged so it can be included in the next version, you'll probably have to merge the changes in the changelog

projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
```diff
@@ -3,29 +3,45 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { AfterViewInit, Directive, ElementRef, HostBinding } from '@angular/core';
+import { AfterViewChecked, Directive, ElementRef, HostBinding } from '@angular/core';
+
+declare type colSize = '2' | '4' | '8' | '10';
```
juanmendes, line 8: Types are usually `TitleCase`

`declare type ColSize = ...`

juanmendes (approved):
Please remove the change to package.json, I will bump it separately including the three changes that are going in soon

PR #274, opened 2020-12-09
[a11y] Fix no alternative text on input error icons

georgiatanasov147:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
- Introduces alternative-text directive that sets the 'alt' attribute on the inner 'svg'

- on the examples, inspected the error icon and verified that it has an empty 'alt' attribute

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/components/src/lib/directives/alternative-text/alternative-text.directive.spec.ts
+++ b/projects/components/src/lib/directives/alternative-text/alternative-text.directive.spec.ts
@@ -1,0 +1,66 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { Component } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ClarityModule } from '@clr/angular';
+import { WidgetObject } from '../../../utils/test';
+import { ALTERNATIVE_TEXT } from './alternative-text.directive';
+import { AlternativeTextModule } from './alternative-text.module';
+
+interface TestContext {
+    widget: ComponentTestWidget;
+    fixture: ComponentFixture<TestHostComponent>;
+}
+
+@Component({
+    template: `
+        <clr-icon shape="home" [vcdAlternativeText]="altText">
+            <svg></svg>
+        </clr-icon>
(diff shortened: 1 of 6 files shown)


projects/components/src/lib/directives/alternative-text/alternative-text.directive.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+class ComponentTestWidget extends WidgetObject<TestHostComponent> {
+    public getIconSVGAlternativeText(shape: string): string {
+        return this.findElement(`clr-icon[shape="${shape}"] svg`).nativeElement.getAttribute('alt');
+    }
+}
+
+describe('AlternativeTextDirective', () => {
+    let widget: ComponentTestWidget;
```
juanmendes, line 36: We don't use closure variables in our tests, we use test objects. This way, you don't need to nullify references

projects/components/src/lib/directives/alternative-text/alternative-text.directive.ts
```diff
@@ -0,0 +1,27 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { AfterViewInit, Directive, ElementRef, Input, Renderer2 } from '@angular/core';
+
+const ALTERNATIVE_TEXT = 'alt';
+
+/**
```
juanmendes, line 10: Why would this be for clr-icon only?

And what's the value of this besides just adding the alt attribute?

georgiatanasov147: Well, my thoughts for the directive were: because I need to set it to the inner svg which is injected inside the clr-icon and that I need to do the same in more than one place.

juanmendes: I see, I didn't realize we're adding the tag to something we don't render ourselves.

juanmendes (approved):

PR #278, opened 2020-12-10
[a11y] Fix signpost bugs in form components 

georgiatanasov147:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature
-   [x] Refactoring (no functional changes, no api changes)
-   [ ] Build related changes
-   [ ] CI related changes
-   [x] Example website changes
-   [ ] Version bump

## What does this change do?
- Added a single line style which sets the following behavior:
  - Sequences of whitespace will collapse into a single whitespace.
  - Text will wrap when necessary, and on line breaks.

- Updated the vcd-form-input example with a longer, more complex signpost text

- Observed that the signpost text in the vcd-form-input example wraps on line breaks,
when necessary and collapses multiple white spaces into a single whitespace

image
image
image

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/components/src/form/form-input/form-input.component.scss
+++ b/projects/components/src/form/form-input/form-input.component.scss
@@ -14,7 +14,3 @@
 .form-group-description {
     margin-bottom: 15px;
 }
-
-clr-signpost p:first-child {
-    margin-top: 0;
-}
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Accessibility: Added blank alternative text to all svgs inside clr-icon elements
 - Label text is not wrapped in `vcd-select-form`
 - Labels is not associated with the 'Access Level' combo box.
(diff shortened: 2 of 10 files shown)


projects/components/src/form/form-input/form-input.component.scss
```diff
@@ -17,4 +17,8 @@
 
 clr-signpost p:first-child {
     margin-top: 0;
+
+    // Sequences of whitespace will collapse into a single whitespace.
+    // Text will wrap when necessary, and on line breaks.
+    white-space: pre-line;
```
juanmendes, line 23: This is not the same as what we use in https://github.com/vmware/vmware-cloud-director-ui-components/blob/a11y/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.scss#L1 Can we make them be the same? Should we use a separate CSS file that is included by both?


Shouldn't we fix this on form-checkbox which also has a signpost? Please see https://github.com/vmware/vmware-cloud-director-ui-components/blob/a11y/projects/components/src/form/form-checkbox/form-checkbox.component.html#L21

juanmendes (approved):

PR #287, opened 2020-12-16
Allow disabling responsive input directive on form controls

juanmendes:

## PR Checklist

## PR Type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [x] Version bump
-   [x] Unit Tests added

## What does this change do?
Allow users of form controls to disable the behavior that makes labels be on the same line when there's enough real estate.

This is needed when the user already has added clr-vertical-form which means they always want the form to be vertical (label and input on different lines)

* Make sure the screen is really wide (> 1400 px wide)
* Loaded the temp build onto VCD_UI
* Started the tenant dev environment
* Went to both vapps and VMs advanced filtering (VDC specific or the global one)
* Opened the advanced filter
* Made sure labels were not wrapping
  * They were wrapping previously because labels were being set to `clr-col-md-2`
* Made sure inputs were taking up the full width
  * Previously they had `clr-col-md-10` which made it not take up the full width 
* Made sure that the responsive grid styles were not being added when the form controls have `[isResponsive]="false"` 

image

image

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
+++ b/projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
@@ -3,9 +3,13 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { AfterViewChecked, Directive, ElementRef, HostBinding } from '@angular/core';
+import { AfterViewChecked, Directive, ElementRef, HostBinding, Input } from '@angular/core';
 
-declare type ColSize = '2' | '4' | '8' | '10';
+type ColSize = '2' | '4' | '8' | '10';
+
+export interface ResponsiveInputOptions {
+    disabled: boolean;
+}
 
 /**
  * Adds Clarity grid classes to form controls based on the host's width so that labels and inputs are on separate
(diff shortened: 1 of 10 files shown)


projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
```diff
@@ -21,10 +25,29 @@ export class ResponsiveInputDirective implements AfterViewChecked {
...
+    }
+
+    private _disabled = false;
+
     @HostBinding('class.clr-row') public clrGridRow = true;
+
     constructor(private el: ElementRef<HTMLElement>) {}
 
     ngAfterViewChecked(): void {
+        if (this._disabled) {
```
yumengwu, line 48: Why not use one `if` statement here like:
`if (this._disabled || this.elementWidth !== 0)`

yumengwu (approved):

projects/components/src/lib/directives/responsive-input/responsive-input.directive.ts
juanmendes: I think it's simpler to see multiple early returns

It's clearer in which cases the function does nothing. Otherwise you may need a gigantic comment explaining your boolean expression

PR #307, opened 2021-02-03
Update Readme

ps37:
Add a table showing Components package versions corresponding to VMware cloud director(VCD) application releases

## PR Checklist

-   [ N/A] Tests for the changes have been added (for bug fixes/features)
-   [N/A ] Examples have been added / updated (for bug fixes / features)
-   [N/A ] Changelog has been updated

## PR Type

-   [ ] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ X ] Documentation content changes
-   [ ] Version bump

## What does this change do?
Update readme


README.md
```diff
@@ -99,3 +99,10 @@ To get further help, please file issues on github.
 
 To get more help on the Angular CLI use `ng help` or go check out the
 [Angular CLI README](https://github.com/angular/angular-cli/blob/master/README.md).
+
+## Components package versions corresponding to VMware cloud director(VCD) application releases
+
```
ryan-bradford, line 104: Should we also put the version of Clarity we require? I know our package.json in components isn't perfectly up to date.

juanmendes: We should and we should fix the package.json to have the correct dependencies

ps37: Following are my thoughts:
- As our components use `Clarity`, we should list `@clr/` packages in the peer dependencies list of our components project
- We should then also fix the versions of `@angular/` packages listed there. They should be updated to angular versions with which the above-listed Clarity version works
- If we do the first thing above, I don't think we need to list Clarity in the version matrix

PR #315, opened 2021-03-01
Adding Widget Object Interaction

juanmendes:
## Allowing Widget Objects to know something about the elements they wrap

The first generation of widget object had a single purpose:
*  Finding references to HTML that a component manages to avoid CSS queries being duplicated.

That restriction made it so that calling code (jasmine, cypress) needed to repeat steps when an interaction was needed, causing duplication and adding verbosity to the spec files.

In more technical terms, the `WidgetObject` didn't have any knowledge of the type of objects it found (it was simply `T`) so a widget object implementation could not write a `login(user, pwd)` method. This changelist introduces a new interface `ElementActions` that allows widget object implementations to interact with the elements it finds through `WidgetObjectElement` which implements `ElementsActions` and `LocatorDriver`.

Implementations of testing environments (such as the provided ones: cypress and angular) now need to implement the `ElementsActions` interface. This is very limited currently and can only click and type into objects. This is enough for  80% of use cases but we will be implementing a fuller interface very soon.

### Testing
This has been tested through the existing data exporter e2e test by adding an `exportData(columnsToExclude)` method. It was previously there just to test the widget object pattern but now has been improved to do something useful:
* Ensuring the file download works
* Ensuring the file's contents are accurate
* Ensuring the dialog is dismissed after the download finishes

data-exporter

## Widget Objects don't read the DOM
There is no interface to read text or DOM attributes. This is intentional because we don't want widget objects making assertions, those should be done by callers on the exposed elements. 

## WidgetObjectElement
 `LocatorDriver` has been renamed to be WidgetObjectElement which implements the LocatorDriver interface.

`WidgetObjectElement`s should be returned only in private methods, they should not be exposed to tests, they are only used
when creating methods that abstract interactions. For the public interface, always return `<T>`

### Creating Element accessors

* `private _getOkButton =  this.factory._dataUi(Data.okButton) // accessor returns WidgetObjectElement` 
* `public getUsernameField = this.factory.dataUi(Data.usernameField) // accessor returns `<T>`

If you need both a public and a private version, you can do the following
 * `private _getOkButton =  this.factory._dataUi(Data.okButton) // accessor returns WidgetObjectElement` 
* `public getOkButton = this.factory.unwrap(this._getOkButton) // accessor returns `<T>`

## Data-UI Attributes
We are suggesting that data-ui HTML attributes be used as hooks for anything tests need to interact with. Also, the information about which data-ui attributes exist shall be exposed in an `export const DataUi = {okButton: 'ok-btn', 'cancel-btn'}`;

This has various benefits:
* Minimizes chances of style or layout related changes breaking tests since there's no sharing of semantic data between the tests and CSS/HTML besides the custom data attribute.
* Makes it easier to strip them in production
* Allows use to write tools that generate starter widget objects that know how to access all its data-ui attributes.
* Minimizes duplication of queries since the `WidgetObject` can import the `DataUi` object, which is the contract between the component and its `WidgetObject`
* Auto-completion when adding the attributes into the HTML

### Testing
The data exporter has been updated not to rely on any internal HTML or CSS besides the DataUi attributes. Its unit tests have been updated heavily.

## PR Checklist

## PR Type
-  [x] Refactoring (no functional changes, with breaking api changes)

-   [x] Yes

I will be fixing vcd_ui's implementation because of these breaking changes as my next task.


projects/components/src/data-exporter/data-exporter.component.spec.ts
```diff
@@ -66,19 +66,11 @@ describe('DataExporterColumnsWithoutDisplayName', () => {
...
-        expect(dropdownMenuItemTexts.text()).toEqual('col1');
-
-        exporter.getColumnCheckbox(0).click();
-
-        const columnBubbles = exporter.getColumnBubbles();
-        exporter.getColumnDropdown().click();
-
-        expect(exporter.getColumnBubbles().text()).toEqual('col2');
+        expect(columnBubbles).toEqual(expects);
+        expect(columnCheckboxLabels).toEqual(expects);
```
juanmendes, line 73: @ryan-bradford  I couldn't understand what this was doing previously, it seemed unrelated to what this test is for. Should I create another test?

ryan-bradford: I think I patched this onto the test to get coverage for removing the column VIA the bubbles. Making another test would probably be good.

cypress/integration/find-cypress-widget.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+export function findCypressWidget<W extends BaseWidgetObject<Chainable>>(
+    widgetConstructor: FindableWidget<Chainable, W>,
+    findOptions?: FindCypressWidgetOptions
+): W {
+    return new CypressWidgetObjectFinder<Chainable>().find(widgetConstructor, findOptions) as W;
+}
+
+export function findDataExporter(options?: FindCypressWidgetOptions): DataExporterWidgetObject<Chainable> {
```
ryan-bradford, line 34: Are we saying as a pattern, all the find{WidgetName}s are going to live in the find-cypress-widget.ts file?

projects/components/src/data-exporter/data-exporter.component.spec.ts
```diff
@@ -122,34 +114,27 @@ describe('VcdExportTableComponent', () => {
                     .toArray()
                     .map((it) => it.text())
             ).toEqual(['Name', 'Description']);
-            exporter.getColumnCheckbox(0).click();
+            exporter.getColumnCheckbox().toArray()[0].click();
```
ryan-bradford, line 117: Should we make this `getColumnCheckboxes()`?

projects/components/src/utils/test/widget-object/angular/angular-widget-object-element.ts
```diff
@@ -1,66 +1,82 @@
...
-            this.rootElement
+            )
         );
     }
 
+    click(): void {
+        this.testElement.click();
+    }
+
+    check(options?: unknown): void {}
```
ryan-bradford, line 55: NIT: Can we add the 

```
/**
 * @inheritdoc
 */
```

Just for neatness?

projects/components/src/utils/test/widget-object/angular/angular-widget-object-element.ts
```diff
@@ -95,48 +114,60 @@ export class AngularLocatorDriver implements LocatorDriver<TestElement> {
 export class TestElement implements Iterable<TestElement> {
     constructor(public elements: DebugElement[], public fixture: ComponentFixture<any>) {}
 
+    private get firstNativeElement(): HTMLElement {
+        return this.firstDebugElement.nativeElement;
```
ryan-bradford, line 118: Should this or firstDebugElement throw an error if there's more than one element present? Might catch an error where there's a query that returns multiple elements but the tester thought there should only be one

projects/components/src/data-exporter/data-exporter.component.spec.ts
juanmendes: Thanks, I'll add its own it for that. Let's avoid covering two problems in one unit test.

projects/components/src/utils/test/widget-object/angular/angular-widget-object-element.ts
```diff
@@ -161,6 +192,13 @@ export class TestElement implements Iterable<TestElement> {
         return this.elements.map((el) => new TestElement([el], this.fixture));
     }
 
+    /**
+     * Maps over each element of this collection
+     */
+    map<U>(callbackfn: (el: TestElement, index: number, array: TestElement[]) => U): U[] {
```
ryan-bradford, line 198: Total extra => is it possible for us to define 
`map = this.toArray().map` and then we'll get all the typescript info + the signature of map for free?

projects/components/src/utils/test/widget-object/cypress/cypress-widget-object-element.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    unwrap(): T {
+        return this.chainable;
+    }
+
+    click(options?: unknown): void {
```
ryan-bradford, line 56: NIT: Same with the  `@inheritdoc` as above?

projects/components/src/utils/test/widget-object/widget-object.ts
```diff
@@ -2,111 +2,220 @@
...
+ * Actions that can be taken on elements from within widget objects. This
+ * is a subset of the functionality from Cypress
  */
-export interface LocatorDriver<T> {
-    /**
-     * Finds all child elements that match the given cssSelector.
-     */
-    get(cssSelector: string, options?: unknown): LocatorDriver<T>;
+export interface ElementActions {
+    click(options?: FindElementOptions): void;
```
ryan-bradford, line 47: NIT: Can we give these methods comments?

projects/components/src/utils/test/widget-object/widget-object.ts
```diff
@@ -2,111 +2,220 @@
...
+     *
+     *   Each factory method returns an ElementLocator<WidgetObjectElement<T>> which should never be public
+     *   but also has a corresponding method prefixed with _ to indicate the it returns `T` which is what should be
+     *   consumed by users of the class.
+     */
+    protected factory = {
+        /**
+         * Utility to create versions of factories that return the unwrapped T
+         */
+        unwrap: (fun: ElementLocator<WidgetObjectElement<T>>) => {
```
ryan-bradford, line 160: I think this is useful to leave here but I'm not sure anyone is really going to use it. I think most people will end up creating the duplication. 

But a bigger question is how do we prevent people from exposing WidgetObjectElements from their widget objects?

Also, I'm not sure the names `_css` and `css` make it totally clear what's the difference between the two besides one's an internal call? Can you rename `_css` to `_cssWidgetInternal`? This makes it clear that it's returning a widget and extra clear that it should be only used internally?

projects/examples/src/components/data-exporter/data-exporter.example.component.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright 2019-2020 VMware, Inc.
  * SPDX-License-Identifier: BSD-2-Clause
  */
-
+//
```
ryan-bradford, line 5: NIT: delete these slashes?

tsconfig.json
```diff
@@ -20,7 +20,7 @@
     ],
     "paths": {
       "@vcd/ui-components": [
-        "dist/components"
+        "projects/components/src"
```
ryan-bradford, line 23: Undo this.

ryan-bradford (changes requested):

PR #317, opened 2021-03-04
Added sections and filters to quick search

ryan-bradford:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?

Improves quick search in the following ways:
1. Adds section that are groups of providers. Sections display as headers with smaller provider headers within them.
2. Adds filters that can dynamically be added/removed and filter the contents of quick search.

1. Loaded into VCD_UI and implemented global search using this new feature.

quick+global

## Very Zoomed

Screen Shot 2021-03-16 at 1 43 05 PM
Screen Shot 2021-03-16 at 1 43 10 PM
Screen Shot 2021-03-16 at 1 43 18 PM

## Questions
1. Should we provide some default type filter or is that pointless?

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.html
+++ b/projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.html
@@ -1,0 +1,16 @@
+Click <button class="btn" (click)="spotlightOpen = true">here</button> to open the Quick Search. <br /><br />
+Quick search gives the client the ability to register filters to filter search results. A filer simply has a name and
+options that can be selected. These filters are activated by typing name:option.
+<br /><br />
+A filter can then filter either the list of providers that are active, or filter the results in a given provider. This
+is all handled from the provider side.
+<br /><br />
+If two filters of the same name are present, they should be joined by "or". If two filters of different names are
+present, they should be joined by "and".
+<br /><br />
+In this example, there are two filters present: "type" and "is". Type will filter the results by provider ID, and is a
+dynamic filter that will filter the results on if they are real action.
+<br /><br />
+Users are able to navigate through the filters both by typing and by using the mouse. Typing (name): will bring up a
+dropdown that allows the user to either scroll through using the keyboard or by clicking.
+<vcd-quick-search [(open)]="spotlightOpen" [filters]="filters"> </vcd-quick-search>
--- a/projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.ts
+++ b/projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.ts
@@ -1,0 +1,105 @@
+/*!
+ * Copyright 2020 VMware, Inc.
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import {
+    ActiveQuickSearchFilter,
+    QuickSearchFilter,
+    QuickSearchProviderDefaults,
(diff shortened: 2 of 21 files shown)


tsconfig.json
```diff
@@ -20,7 +20,7 @@
     ],
     "paths": {
       "@vcd/ui-components": [
-        "dist/components"
+        "projects/components/src"
```
ryan-bradford, line 23: Undo this.

projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.html
```diff
@@ -0,0 +1,22 @@
+Press <kbd>Command</kbd>+<kbd>f</kbd> on MacOS / <kbd>Ctrl</kbd>+<kbd>f</kbd> on PC or click
```
juanmendes, line 1: I don't see anything here referring to explaining to use the filter icon and that it will filter for results of that service provider type. 

Is this copy pasted from another example? Why is there an input with a placeholder?

Please rework thinking about a user that has never seen the feature before and is coming to this specific examples

juanmendes: Make sure we explain both ways, using the mouse, or just the keyboard. We want to explain the feature in these examples

projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.ts
```diff
@@ -0,0 +1,174 @@
...
+    ];
+
+    private lazyLoadedProvider = new LazyLoadedActionsSearchProvider();
+    private actionsSearchProvider = new ActionsSearchProvider();
+
+    private readonly mousetrap: MousetrapInstance;
+    constructor(private fb: FormBuilder, private searchRegistrar: QuickSearchRegistrarService) {
+        // Create an instance of mouse trap within the constructor so that any bount shortcut event handler
+        // would be executed within the angular zone
+        this.mousetrap = new Mousetrap();
```
juanmendes, line 42: Do we really need mouse trap? We should avoid libraries in the examples, they should  be as simple as possible. It was a mistake to use it before.

It's very simple to set a key handler with a modifier with angular

```<input (keydown.meta.f)="onKeydown($event)">```

File: projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.ts [hunk 4]
```diff
@@ -0,0 +1,174 @@
+/*!
```
juanmendes, line 1: Did you create this using the generator created by Yumin? Or copy pasted from another example? Please be sure to always use the generator, we don't want remnants of other tests.

juanmendes:
We also need an example showing nested providers...

projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.ts
ryan-bradford: Removed mousetrap, it opens with the button now (which is much larger in this example)

projects/examples/src/components/quick-search/filters/quick-search-filters.example.component.ts
ryan-bradford: Will use the generator in the future.

juanmendes (approved):

PR #322, opened 2021-03-19
Fixed datagrid infinite selection loop

ryan-bradford:
## PR Checklist

## PR Type

-   [X] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?

Two way binding has a problem where an event from Clarity, causes the client to send an event back, which causes an infinite loop. 

This change stops the propagation of the events after one iteration.

1. Ran the selection example and observed the changelog spam stopping.

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -296,14 +296,25 @@ describe('DatagridComponent', () => {
                     this.clrGridWidget.getSelectionLabelForRow(1).click();
                     await timeout();
                     expect(this.hostComponent.selectionChanged).toHaveBeenCalledWith(mockData);
-                    await timeout();
+                });
+
+                it('only emits events when selection actually changes', async function (this: HasFinderAndGrid): Promise<
+                    void
+                > {
+                    this.hostComponent.selectionType = GridSelectionType.Multi;
+                    this.finder.detectChanges();
+                    spyOn(this.hostComponent, 'selectionChanged');
+                    this.hostComponent.datagridSelection = [mockData[0]];
                     this.finder.detectChanges();
                     await timeout();
                     this.finder.detectChanges();
+                    expect(this.hostComponent.selectionChanged).toHaveBeenCalledTimes(1);
                     await timeout();
                     this.finder.detectChanges();
+                    expect(this.hostComponent.selectionChanged).toHaveBeenCalledTimes(1);
                     await timeout();
-                    expect(this.hostComponent.selectionChanged).toHaveBeenCalledTimes(3);
+                    this.finder.detectChanges();
+                    expect(this.hostComponent.selectionChanged).toHaveBeenCalledTimes(1);
                 });
 
                 it('emits only one row when set to single selection', function (this: HasFinderAndGrid): void {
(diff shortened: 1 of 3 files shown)


projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -296,6 +296,14 @@ describe('DatagridComponent', () => {
...
                     await timeout();
                     expect(this.hostComponent.selectionChanged).toHaveBeenCalledWith(mockData);
+                    await timeout();
+                    this.finder.detectChanges();
+                    await timeout();
+                    this.finder.detectChanges();
+                    await timeout();
+                    this.finder.detectChanges();
+                    await timeout();
+                    expect(this.hostComponent.selectionChanged).toHaveBeenCalledTimes(3);
```
irahov, line 306: Shouldn't this verify that it has been called just once? Why `selectionChanged` should have been called all the time provided that there is no activity that leads to a real change?

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -827,9 +827,11 @@ export class DatagridComponent<R extends B, B = any> implements OnInit, AfterVie
         if (this.datagrid.selection.currentSingle) {
             this._datagridSelection = [this.datagrid.selection.currentSingle];
         } else if (this.datagrid.selection.current && this.datagrid.selection.current.length) {
-            this._datagridSelection = [...this.datagrid.selection.current];
+            this._datagridSelection = this.datagrid.selection.current;
         } else {
```
irahov, line 831: I would write this in the pattern of the previous rows, i.e.:
} else if (this._datagridSelection.length) {
}

and add a comment - that if there is no selection from the clarity we should clear our (in case there is such) without creating a new reference.

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -827,9 +827,11 @@ export class DatagridComponent<R extends B, B = any> implements OnInit, AfterVie
         if (this.datagrid.selection.currentSingle) {
```
irahov, line 827: Shouldn't we change this one also?
if (this.datagrid.selection.currentSingle) {
   if (this._datagridSelection?.length !== 1 || this._datagridSelection[0] !== this.datagrid.selection.currentSingle) {
      this._datagridSelection = [this.datagrid.selection.currentSingle];
   }
}

irahov (approved):

projects/components/src/datagrid/datagrid.component.spec.ts
ryan-bradford: I think this is the correct behavior, but it's unclear because I tacked this onto an existing test, which was a bad decision. Moving it to a new `it`

projects/components/src/datagrid/datagrid.component.ts
ryan-bradford: Why write it in the pattern of the previous rows? If we do that, once the selection length gets to 0, it would be impossible to select another entity, right?

projects/components/src/datagrid/datagrid.component.ts
ryan-bradford: 👍

PR #326, opened 2021-03-29
Update the action availability to accept observables

ps37:
## PR Checklist

## PR Type

-   [X] Bugfix
-   [X] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
Some of the actions, particularly the visibility static actions depend
on values other than selected entities, like the response from async
requests. For such cases, we need to accept observables to keep their
visibility up to date.
There are some bugs in the VCD ui related to the visibility of static
actions not getting updated because the visibility is not calculated
when one of the variables on which their visibility depends is not
updated.

- Started the local dev server on localhost:4200
- Opened the "Hiding and disabling of actions" section of the "Action Menu" tab in the examples app
- Made sure that the "Action availability as observable" section is visible
- In that section, making sure that the visibility of static featured "Action 1" and the contextual "Nested Action 1" is toggled every 2 seconds
- Also, visited the "Static and contextual actions" section made sure that the examples are shown the same as what is being rendered in https://vmware.github.io/vmware-cloud-director-ui-components/action-menu/example/action-menu-static-and-contextual-actions
- Ran the unit tests and made sure that all the unit tests are passing
- Changed the availability of the static actions to observable in the following VCD UI screens and tested that they work as expected:
1. Trusted certificates
2. Certificates library
3. Edge gateways
4. Users
5. External networks list

https://user-images.githubusercontent.com/10735165/112892409-436bab80-90a7-11eb-9e26-48ea4ec60190.mov

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/components/src/action-menu/action-menu.component.ts
+++ b/projects/components/src/action-menu/action-menu.component.ts
@@ -4,6 +4,7 @@
  */
 
 import { Component, EventEmitter, Input, Output, TrackByFunction } from '@angular/core';
+import { isObservable } from 'rxjs';
 import { ActionDisplayConfig, ActionItem, ActionStyling, ActionType, TextIcon } from '../common/interfaces';
 import { CommonUtil } from '../utils';
 
@@ -50,16 +51,6 @@ export class ActionMenuComponent<R, T> {
         return this._actions;
     }
 
-    /**
-     * Used by the {@link EntityActionExtensionComponentsDirective} to provide the extension actions from plugins after
-     * they are created asynchronously
-     * @param val List of extension actions
-     */
-    set extensionEntityActions(val: ActionItem<unknown, unknown>[]) {
-        this._extensionEntityActions = val;
-    }
-    private _extensionEntityActions: ActionItem<unknown, unknown>[];
-
     private _actionDisplayConfig: ActionDisplayConfig = getDefaultActionDisplayConfig();
     /**
      * Display configuration of both static and contextual actions
@@ -76,6 +67,12 @@ export class ActionMenuComponent<R, T> {
     get actionDisplayConfig(): ActionDisplayConfig {
         return this._actionDisplayConfig;
     }
+
+    /**
+     * Copy of actions passed with their availability call backs. This is because, when the selected entities get updated, we need to use
+     * those CBs to calculate the availability again
+     */
+    private actionsWithAvailabilityCb: ActionItem<R, T>[];
 
     /**
      * When there are no nested actions and if all of the contextual actions are marked to be featured, there is no need to show
@@ -111,7 +108,7 @@ export class ActionMenuComponent<R, T> {
      */
     @Input() set selectedEntities(val: R[]) {
         this._selectedEntities = val;
-        this.updateActionListsAndDisplayFlags();
+        this.updateDisplayedActions();
     }
     private _selectedEntities: R[] = [];
     get selectedEntities(): R[] {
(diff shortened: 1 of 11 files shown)


projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -77,6 +68,12 @@ export class ActionMenuComponent<R, T> {
         return this._actionDisplayConfig;
     }
 
+    /**
+     * Copy of actions passed with their availability call backs. This is because, when the selected entities get updated, we need to use
+     * those CBs to calculate the availability again
+     */
+    private actionsWithAvailabilityCb: ActionItem<R, T>[];
```
ryan-bradford, line 75: Its confusing that some filtering happens in the template and some happens in the component. There are also a few methods that exist that I'm not sure need to.

I think this would be more clear if the following was true:
1. Action item availability is input as either a boolean, a function, or an observable
2. We convert availability to always be an Observable and store it in this private variable (`actionsWithAvailabilityAsObs`)
3. This removes all calls to `isObservable` in the templates, and cleans up methods like `isActionAvailable`.
4. In dropdown.component.ts, we now require that all availability is input as an observable. This again cleans up the code. 

So pretty much, accepting availability as a boolean or a function is a convenience thing for clients, and we convert everything to an observable. This will clean up the template code, and the typescript code too.

We might want to delay this so we can get the feature in by beta though... if so, let's create a follow up ticket and do it right away

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -254,12 +251,31 @@ export class ActionMenuComponent<R, T> {
...
         this.shouldDisplayContextualActionsDropdownInline =
             hasNestedActions || this._actions.some((action) => action.actionType === ActionType.CONTEXTUAL);
 
         this.updateActionListsAndDisplayFlags();
     }
 
+    /**
+     * Executes the availability call backs and updates them to booleans
+     */
+    updateActionsAvailability(actions) {
```
ryan-bradford, line 265: Rename to something that suggests the changing of callbacks to booleans.

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -288,7 +304,17 @@ export class ActionMenuComponent<R, T> {
      * disabled mode
      */
     private isActionAvailable(action: ActionItem<R, T>): boolean {
-        return !action.availability || action.availability(this.selectedEntities) || this.isActionDisabled(action);
+        let isActionAvailable = true;
```
ryan-bradford, line 307: Add comment or rename for the case that avail is an observable.

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -407,6 +433,13 @@ export class ActionMenuComponent<R, T> {
             this.actionDisplayConfig.contextual.styling === style
         );
     }
+
+    /**
+     * Used for deciding if the availability has to be passed through an Async pipe in the template
+     */
+    isObservable(availability): boolean {
```
ryan-bradford, line 440: NIT: `isObservable = isObservable`

ryan-bradford (approved):

projects/components/src/action-menu/action-menu.component.ts
ps37: > We convert everything to an observable

As discussed with Ryan on zoom call, I think this is going to be problematic for the following reasons:
- We have to create new observables everytime the callbacks/booleans get updated because of changes in selected entities or actions list
- The calculation of availability is required on the TS side for UX things like:
1. Ungrouping single nested actions
2. The number of featured actions shown should be equal to the featured count. For example, In the following scenario where Featured actions: 3, Featured count: 2, If the 2nd featured action is unavailable, and we don't filter that, we only show the first action in featured bar!
- So, the code calling available actions in the class should become async and this makes the collection(action list) processing complicated. And I don't see this as necessary right now.

If we still want to explore this approach, I created a task to do it later: https://jira.eng.vmware.com/browse/VDUCC-615

PR #331, opened 2021-04-06
Allows findWidget to use all given findOptions and adds clear to widget object

ryan-bradford:
## PR Checklist

## PR Type

-   [X] Bugfix
-   [X] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
The widget finder was ignoring most of the arguments given to `findOptions` in a call to `findWidget`. This meant we couldn't find widgets by text or index. This patches the bug.

Also, this adds tests for the cypress widget objects. It does this by mocking the cy object and making assertions about calls to it.

Furthermore, clear was not a method that was callable from inside a widget object, even though it should have been. This adds clear as a method to a widget object.

1. Ran the data exporter tests and observed them passing.

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
+++ b/projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
@@ -40,11 +40,23 @@ export class CypressWidgetObjectFinder<T> {
         if (findOptions?.cssSelector) {
             tagName += `${findOptions.cssSelector}`;
         }
+
         const id = idGenerator.generate();
-        const search = findOptions?.ancestor
-            ? cy.get(findOptions.ancestor).find(tagName)
-            : cy.get(tagName, findOptions?.options);
-        const root = search.as(id);
+
+        const ancestor = findOptions?.ancestor ? cy.get(findOptions?.ancestor) : cy.get('body');
+        const parentWidget = new CypressWidgetObjectElement(ancestor, false, undefined);
+        let query = widgetConstructor.tagName;
+        if (findOptions?.cssSelector) {
+            query = query + findOptions.cssSelector;
+        }
+        const parentQuery: FindCypressWidgetOptions = {
+            cssSelector: query,
+            text: findOptions?.text,
+            index: findOptions?.index,
+            options: findOptions?.options,
+        };
+
+        const root = parentWidget.get(parentQuery).unwrap().as(id);
         return new widgetConstructor(new CypressWidgetObjectElement(root, true, id));
     }
 }
--- a/projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
+++ b/projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
@@ -43,7 +43,9 @@ export class CypressWidgetObjectFinder<T> {
 
         const id = idGenerator.generate();
 
-        const ancestor = findOptions?.ancestor ? cy.get(findOptions?.ancestor) : cy.get('body');
+        const ancestor = findOptions?.ancestor
+            ? cy.get(findOptions?.ancestor, { timeout: findOptions?.options?.timeout })
+            : cy.get('body', { timeout: findOptions?.options?.timeout });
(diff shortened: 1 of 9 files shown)


projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
```diff
@@ -40,11 +40,23 @@ export class CypressWidgetObjectFinder<T> {
         if (findOptions?.cssSelector) {
             tagName += `${findOptions.cssSelector}`;
```
xvu-vmware, line 41: Don't you need to prefix with a space if you're going to concat the two selectors?

For example:
```tagName += ` ${findOptions.cssSelector}`;```

File: projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts [hunk 2]
```diff
@@ -40,11 +40,23 @@ export class CypressWidgetObjectFinder<T> {
...
         }
+
         const id = idGenerator.generate();
-        const search = findOptions?.ancestor
-            ? cy.get(findOptions.ancestor).find(tagName)
-            : cy.get(tagName, findOptions?.options);
-        const root = search.as(id);
+
+        const ancestor = findOptions?.ancestor ? cy.get(findOptions?.ancestor) : cy.get('body');
```
xvu-vmware, line 46: We're not using the `FindCypressWidgetOptions.options` for timeout.

For example:
```
// This code isn't complete. Don't pass in `timeout` if it's not defined.
const timeout = findOptions?.options?.timeout;
const ancestor = findOptions?.ancestor ? cy.get(findOptions?.ancestor, { timeout } ) : cy.get('body',  { timeout });
```

projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
```diff
@@ -40,11 +40,23 @@ export class CypressWidgetObjectFinder<T> {
...
+        const parentWidget = new CypressWidgetObjectElement(ancestor, false, undefined);
+        let query = widgetConstructor.tagName;
+        if (findOptions?.cssSelector) {
+            query = query + findOptions.cssSelector;
```
xvu-vmware, line 50: Don't you need a space when combining selectors?

xvu-vmware (approved):

projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
juanmendes: No, the selector should be applied to the root element, it's combining it with the tag name? I believe, @ryan-bradford ?

juanmendes:
It doesn't make sense to search for a widget by text, right? Can we remove that option?

xvu-vmware:
> It doesn't make sense to search for a widget by text, right? Can we remove that option?

Imagine I'm on the vApp card screen and I would like to perform power on action for "vApp-1". I want to call `findWidget()` to get the vApp card widget that has the vApp name of "vApp-1". I think it makes sense to be able to find by text.

projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
ryan-bradford: Yea, there's a case where you want to get some tagName with a specific class applied (`vcd-data-exporter.vms-exporter`), and for this case, we can't have a space.

projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
ryan-bradford: Same as in the below discussion.

projects/components/src/utils/test/widget-object/cypress/cypress-widget-finder.ts
ryan-bradford: Updated to include this!

PR #332, opened 2021-04-07
Change Availability of static actions to not accept CBs

ps37:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
Callbacks with selected entities are only meant for contextual actions and users were using them for static actions also. So now with this change, when a user tries to assign values to the availability of static actions other than observables, they get an error.

- Fixed the components of examples project which are wrongly using the availability callbacks
- Started the examples app and made sure that the examples render as before
- Copy pasted the code to node modules of vcd ui and noticed that the code in fact shows an error when a callback is assigned to availability
of static actions

Screen Shot 2021-04-07 at 8 14 37 AM
Screen Shot 2021-04-07 at 8 13 20 AM

-   [X] Yes
-   [ ] No

Any existing static actions in VCD ui that use callbacks for their availability are now going to break. As a follow up to this task, I am going to fix all those places in the VCD ui

### DIFF
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## [2.0.0-dev.9]
-### Changed
+### Breaking
 - Changed the interface action availability such that the availability of static actions does not accept call backs
 
 ## [2.0.0-dev.8]
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [2.0.0-dev.8]
+## [2.0.0-dev.9]
+### Changed
+- Changed the interface action availability such that the availability of static actions does not accept call backs
 
+## [2.0.0-dev.8]
 ### Added
 - Find widget by data ui attribute.
 
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## [2.0.0-dev.9]
-### Breaking
-- Changed the interface action availability such that the availability of static actions does not accept call backs
+### Changed
+- BREAKING: Changed the interface action availability such that the availability of static actions does not accept call backs
 
 ## [2.0.0-dev.8]
 ### Added
--- a/projects/components/src/action-menu/action-menu.component.ts
+++ b/projects/components/src/action-menu/action-menu.component.ts
@@ -4,8 +4,15 @@
  */
 
 import { Component, EventEmitter, Input, Output, TrackByFunction } from '@angular/core';
-import { isObservable } from 'rxjs';
-import { ActionDisplayConfig, ActionItem, ActionStyling, ActionType, TextIcon } from '../common/interfaces';
+import { isObservable, Observable } from 'rxjs';
+import {
+    ActionDisplayConfig,
+    ActionItem,
+    ActionStyling,
+    ActionType,
+    BaseActionItem,
+    TextIcon,
+} from '../common/interfaces';
 import { CommonUtil } from '../utils';
 
 /**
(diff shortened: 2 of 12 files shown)


projects/components/CHANGELOG.MD
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 - Added functionality to disable an option in `FormSelectComponent`
 
+## [2.0.0-dev.8]
+### Changed
+- Changed the interface action availability such that the availability of static actions does not accept call backs
```
ryan-bradford, line 13: Let's label this with breaking

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -23,6 +30,21 @@ export function getDefaultActionDisplayConfig(cfg: ActionDisplayConfig = {}): Ac
...
+ */
+interface ActionItemInternal<R, T> extends BaseActionItem<R, T> {
+    /**
+     * Used for determining where in the action menu this action gets displayed
+     */
+    actionType?: ActionType;
+    /**
+     * Condition whether or not the action is available.
+     */
+    availability?: ((selectedEntities: R[]) => boolean) | Observable<boolean> | boolean;
```
ryan-bradford, line 45: I thought we decided that `((selectedEntities: R[]) => boolean)` shouldn't be in the internal availability?

To get around the children problem, you might want to try redeclaring `children` in this interface with the type `ActionItemInternal`.

projects/examples/src/app/components/action-menu/with-separators/action-menu-with-separators.example.component.html
```diff
@@ -12,6 +12,7 @@
         [actions]="actions"
         [actionDisplayConfig]="actionDisplayConfig"
         [dropdownTriggerBtnText]="'vcd.cc.action.menu.actions'"
+        [selectedEntities]="[{}]"
```
ryan-bradford, line 15: Why did you need to add this?

ryan-bradford (changes requested):
Can you add screenshots from vcd_ui showing the actions still working?

projects/components/src/action-menu/action-menu.component.ts
ps37: >To get around the children problem, you might want to try redeclaring children in this interface with the type ActionItemInternal.
That is not working as TS complains that action types are incompatible

projects/examples/src/app/components/action-menu/with-separators/action-menu-with-separators.example.component.html
ps37: Without selected entities, contextual actions will not be shown and in this example, we need the contextual actions to show the menu with separators.

projects/components/src/action-menu/action-menu.component.ts
ryan-bradford: I still wanna push back on this. We shouldn't be modifying the definition of `availability` to get some Typescript issue to go away. Having the callback still in this type is confusing and could lead to problems down the line. 

Maybe children need to be given a type of `BaseActionItem`? And then in here they're redeclared to be `ActionItemInternal`? Like, there's some way to rework these types so they pass without any casts or hacks.

ps37: I removed the call back from ActionItemInternal. However, I still required the casts in 2 places. I think that is fine because, at those points in the code, the action children are converted into ActionItemInternal but TS doesn't know that and we don't want to expose children as ActionItemInternal.

projects/components/CHANGELOG.MD
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## [2.0.0-dev.9]
-### Changed
+### Breaking
```
ryan-bradford, line 11: The format we generally use is 
```
### Changed
- BREAKING: Changed the interface action availability such that the availability of static actions does not accept call backs
```

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -34,15 +34,15 @@ export function getDefaultActionDisplayConfig(cfg: ActionDisplayConfig = {}): Ac
  * We internally convert the callbacks to booleans to avoid calling the callbacks all the time from template. However, we don't want to
  * allow callers to assign boolean variables to availability as there is no way to know when those variables can get updated from outside.
  */
-interface ActionItemInternal<R, T> extends BaseActionItem<R, T> {
+export interface ActionItemInternal<R, T> extends BaseActionItem<R, T> {
```
ryan-bradford, line 37: Why export this?

ryan-bradford (approved):
Small comments + add screenshots from VCD_UI

PR #337, opened 2021-04-13
Improve the action display config interface for datagrid

ps37:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
This change does the following:
- ActionDisplayConfig of ActionMenuComponent does not accept `featuredCount` when actions are displayed inline as there is no difference between featured and non-featured actions in this case
- ActionDisplayConfig of contextual actions of DatagridComponent now takes a new property called `position` which is used to display the actions in Datagrid either in a row or on top of a grid. This replaces the `@input contextualActionPosition`

Tested the following examples to make sure that the contextual actions are rendered inline without any errors even after removing the `featuredCount` property:
`ActionMenuContextualActionsExampleComponent`
`ActionMenuHideDisableExampleComponent`
`ActionMenuSearchExampleComponent`
`ActionMenuSearchDebounceExampleComponent`
`ActionMenuStaticAndContextualActionsExampleComponent`

Tested the `DatagridLinkExampleComponent` to make sure that the actions are displayed in following configurations:
- In rows and as inline bars
- In rows and as dropdowns
- On top and as inline bar

https://user-images.githubusercontent.com/10735165/114630846-6bebcc00-9c89-11eb-9dcf-7eadb2fa517b.mov

-   [X] Yes
-   [ ] No

### DIFF
--- a/projects/components/src/common/interfaces/action-item.interface.ts
+++ b/projects/components/src/common/interfaces/action-item.interface.ts
@@ -157,16 +157,16 @@ export interface ContextualActionDropdownDisplayConfig {
      *
      * If featuredCount is not set, it will default to all featured actions.
      */
-    featuredCount: number;
+    featuredCount?: number;
     /**
-     * How the featured actions should be displayed
+     * To display actions in a dropdown
      */
     styling: ActionStyling.DROPDOWN;
     /**
      * If the title should be the button label, icon, or both
-     * Defaults to ICON if unset.
+     * Defaults to {@link TextIcon.TEXT} when unset
      */
-    buttonContents: TextIcon;
+    buttonContents?: TextIcon;
 }
 
 /**
@@ -175,7 +175,7 @@ export interface ContextualActionDropdownDisplayConfig {
  */
 export interface ContextualActionInlineDisplayConfig {
     /**
-     * How the featured actions should be displayed
+     * To display actions in a inline horizontal ribbon
      */
     styling: ActionStyling.INLINE;
     /**
@@ -191,11 +191,13 @@ export interface ContextualActionInlineDisplayConfig {
 export interface ActionDisplayConfig {
     /**
      * How the contextual actions list shows up on the screen
+     * If this is not specified, this defaults to { featuredCount: 0, styling: ActionStyling.INLINE, buttonContents: TextIcon.TEXT}
      */
     contextual?: ContextualActionDropdownDisplayConfig | ContextualActionInlineDisplayConfig;
 
     /**
      * How the static actions list shows up on the screen
+     * This defaults to ActionStyling.INLINE
      */
     staticActionStyling?: ActionStyling;
 }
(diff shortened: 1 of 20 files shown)


projects/components/src/common/interfaces/action-item.interface.ts
```diff
@@ -177,3 +144,58 @@ export enum TextIcon {
...
+     * Defaults to ICON if unset.
+     */
+    buttonContents: TextIcon;
+}
+
+/**
+ * This along with {@link ContextualActionDropdownDisplayConfig} is one of the types of
+ * {@link ActionDisplayConfig.contextual}
+ */
+export interface ContextualActionInlineDisplayConfig {
```
ryan-bradford, line 176: Can you add a comment explaining what INLINE and DROPDOWN mean? I was confused about this.

ryan-bradford (approved):
Let's also just run through this in the parking lot tomorrow

projects/components/src/datagrid/interfaces/datagrid-action-display.interface.ts
```diff
@@ -0,0 +1,66 @@
...
+    ContextualActionInlineDisplayConfig,
+    TextIcon,
+} from '../../common/interfaces';
+import { ContextualActionPosition } from '../datagrid.component';
+
+/**
+ * Makes the featured count as required when action menu is displayed as a dropdown and also adds the position
+ * at which the actions have to displayed in a datagrid
+ */
+interface DatagridContextualActionDropdownDisplayConfig extends ContextualActionDropdownDisplayConfig {
```
irahov, line 17: What I don't get is: how position in a row will handle `featuredCount`

projects/components/src/datagrid/interfaces/datagrid-action-display.interface.ts
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * An enum that describes where the contextual buttons should display.
+     */
+    position: ContextualActionPosition;
```
irahov, line 21: Do you think it will be better to rename `ContextualActionPosition` to `DatagridContextualActionPosition`, because  when I read `ContextualActionPosition` I'm thinking contextual action position in general

projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -518,15 +516,16 @@ export class DatagridComponent<R extends B, B = any> implements OnInit, AfterVie
...
     /**
      * How to display the static and contextual actions.
      */
-    @Input() actionDisplayConfig: ActionDisplayConfig;
-
-    /**
-     * Whether to display contextual actions inside the row or on top of the grid
-     */
-    @Input() contextualActionPosition: ContextualActionPosition = ContextualActionPosition.TOP;
+    @Input() set actionDisplayConfig(value: DatagridActionDisplayConfig) {
```
irahov, line 523: Here it is good to denote what are the defaults, so if the developer does not  set it it will easily be seen.

irahov (approved):
Something that will help very much is having an example how all these variations work and look like.
It may be in a different commit, if you prefer

projects/components/src/datagrid/interfaces/datagrid-action-display.interface.ts
ps37: Position in a row does not do anything different about the featured count. The number of featured actions shown is the same whether displayed in a row or outside

PR #338, opened 2021-04-14
Fixes bugs with the number with unit form

ryan-bradford:
## PR Checklist

## PR Type

-   [X] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?

1. Fixes a bug where if the unlimitedValue was 0, it was impossible to set the form as required. This is because null was passed through in `getValue()` outputting 0, which was the same as the unlimtedValue.
2. Fixes a bug where if the default value was unlimited, a unit was never calculated causing errors. 
3. Changes it so if the user types in the unlimited value, when the input is blurred, the checkbox will be checked.

1. Typed in the unlimited value, blurred the input, observed the checkbox get checked
2. Set the unlimitedValue to 0 and observed a required error when I cleared the form.
3. Set the default to unlimited through the form control and observed a calculated unit on unselected the unlimited box.

## GIf

number-with-unit

-   [ ] Yes
-   [X] No

### DIFF
--- a/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.spec.ts
+++ b/projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.spec.ts
@@ -4,7 +4,7 @@
  */
 
 import { Component } from '@angular/core';
-import { fakeAsync, TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { FormControl, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { ClarityModule } from '@clr/angular';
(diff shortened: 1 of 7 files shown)


bryanvb:
Looks good to me.

projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.spec.ts
```diff
@@ -137,6 +137,32 @@ describe('VcdNumberWithUnitFormInputComponent', () => {
             numberWithUnitInput.clickUnlimitedCheckbox();
             expect(numberWithUnitInput.isInputValueFocused).toBe(true, 'Input element should have focus');
         });
+
+        it('will choose a unit if initially unlimited when unchecked', () => {
```
irahov, line 141: not clear isn't it better
will choose a unit when unchecked for initial value unlimited

projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.ts
```diff
@@ -239,6 +239,10 @@ export class NumberWithUnitFormInputComponent extends BaseFormControl implements
         this.unlimitedControlValue = unlimitedChecked;
         this.textInputValue = unlimitedChecked ? '' : this.lastRealValue?.toString() || '';
 
+        if (!this.unitsControlValue && !unlimitedChecked) {
```
irahov, line 242: Some comment will help understand why in this case we should set the value by getting the value

projects/components/src/form/number-with-unit-input/number-with-unit-form-input.component.ts
```diff
@@ -333,6 +337,15 @@ export class NumberWithUnitFormInputComponent extends BaseFormControl implements
         }
     }
 
+    /**
+     * Checks if the value in the form is the unlimited value.
```
irahov, line 341: It is confusing:
`Checks if the value in the form is the unlimited value` this means the method should return `true` or `false`

projects/examples/src/components/form-input/number-with-unit-form-input.example.component.ts
```diff
@@ -52,8 +52,8 @@ export class NumberWithUnitFormInputExampleComponent implements AfterViewInit, O
         this.formGroup = fb.group({
             readonly: new FormControl(false),
             disabled: new FormControl(false),
-            memory: new FormControl(1024 * 2, [Validators.required, memoryValidator]),
```
irahov, line 55: Why did you remove the `memoryValidator` ?

projects/examples/src/components/form-input/number-with-unit-form-input.example.component.ts
```diff
@@ -67,9 +67,5 @@ export class NumberWithUnitFormInputExampleComponent implements AfterViewInit, O
         });
     }
 
-    ngAfterViewInit(): void {
-        this.formGroup.controls.cpuLimit.setValue(-1);
```
irahov, line 71: I think there is a case to set the value later and not as part of the formGroup creation. But it is pitty there is no comment (may be it was me that did it and missed to add a comment).
Just double check please

irahov (approved):

projects/examples/src/components/form-input/number-with-unit-form-input.example.component.ts
ryan-bradford: Ok. I'm going to merge this because Bryan needs the fixes. But we can talk about this on Monday in the parking lot.

PR #347, opened 2021-04-21
Optimize ActionMenuComponent for less processing on input change

irahov:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump
-   [x] Other... Please describe: Optimization

## What does this change do?
Currently on any reference input change for `actions` and `selectedEntities`
the component is doing redundant computations and emits `actionsUpdate`
even though there is no actual change.

This input reference change can be as simple as just another empty array `[]`.
There are other cases when the `selectedEntities` are wrongly provided by a getter:
get selectedEntities() {
  return [this.currentSelection];
}
in these cases also there is no actual change, but angular will call the getter
from the template on every digest cycle, the getter will return a new reference
which will lead to performing all those calculations, emitting `actionsUpdate`
and all the subsequent calculations in the corresponding subscribers.

Although it is not good to use a function (the getter in this case) within a template
we are developing a library and cannot control every aspect of how
this library is going to be called that's why we'd like to make it as rigid as possible

For that purpose, this CLN is introducing a shallowEqual comparison for those
two inputs: `actions` and `selectedEntities` and this does quite much of
optimizations

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -6,8 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [2.0.0-dev.16]
 ### Added
 - New unit tests to ensure datagridSelection returns same reference
+### Changed
+- Optimize ActionMenuComponent for less processing on input change
 
 ## [2.0.0-dev.15]
 ### Changed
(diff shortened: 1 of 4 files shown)


ps37:
Can you do manual testing of this in VCD UI app? Especially on the following screens:
- vApp lists and vApp details
- Vm lists and Vm details
- Users list and details
- Edge gateways list and details
- Networks list and details

ps37:
I also think you should bump the package version with this and create a separate MR in Gitlab for Vcd ui using that bumped version. This is so that, we can make sure that vcd ui build is not broken with these changes.

ps37 (approved):

PR #350, opened 2021-04-22
Add unit test to ensure datagridSelection returns same reference

irahov:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump
-   [x] Other... Please describe:

## What does this change do?

If there is no change in the datagrid selection the getter method `datagridSelection`
should return the same array reference. This is needed for performance optimization
so that angular will not trigger any unnecessary change detections and calculations
in the case when the array reference has been changed, but it contains the same data
(for example empty arrays are never equal []!==[])

Ensuring `datagridSelection` returns the same array reference when there is no real selection change
during angular detect changes cycles was done some time ago, but unit tests
were not written. This CLN just adds unit tests for that.

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.spec.ts
+++ b/projects/components/src/datagrid/datagrid.component.spec.ts
@@ -282,6 +282,133 @@ describe('DatagridComponent', () => {
                 });
             });
 
+            describe('datagridSelection getter', () => {
+                describe('when selectionType is set to multi', () => {
+                    beforeEach(function (this: HasFinderAndGrid): void {
+                        this.hostComponent.selectionType = GridSelectionType.Multi;
+                        this.finder.detectChanges();
+                    });
+
+                    it('returns same reference when there is no selection', fakeAsync(function (
+                        this: HasFinderAndGrid
+                    ): void {
+                        const selection = this.hostComponent.grid.datagridSelection;
+                        tick();
+                        this.finder.detectChanges();
+                        expect(this.hostComponent.grid.datagridSelection).toEqual([]);
+                        expect(this.hostComponent.grid.datagridSelection).toBe(selection);
+                    }));
+                    it('returns same reference when there is a selection that was not changed', fakeAsync(function (
+                        this: HasFinderAndGrid
+                    ): void {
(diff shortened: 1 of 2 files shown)


projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -282,6 +282,131 @@ describe('DatagridComponent', () => {
...
+                        const selection = this.hostComponent.grid.datagridSelection;
+                        // Deselect the item
+                        this.clrGridWidget.getSelectionLabelForRow(0).click();
+                        tick();
+                        this.finder.detectChanges();
+                        // Select the same item
+                        this.clrGridWidget.getSelectionLabelForRow(0).click();
+                        tick();
+                        this.finder.detectChanges();
+                        expect(this.hostComponent.grid.datagridSelection).toEqual(selection);
```
ryan-bradford, line 328: Can we expect that this.hostComponent.grid.datagridSelection[0] is the same reference as selection[0]?

projects/components/src/datagrid/datagrid.component.spec.ts
```diff
@@ -282,6 +282,131 @@ describe('DatagridComponent', () => {
...
+                        const selection = this.hostComponent.grid.datagridSelection;
+                        // Select another item
+                        this.clrGridWidget.getSelectionLabelForRow(1).click();
+                        tick();
+                        this.finder.detectChanges();
+                        // Select the first item again
+                        this.clrGridWidget.getSelectionLabelForRow(0).click();
+                        tick();
+                        this.finder.detectChanges();
+                        expect(this.hostComponent.grid.datagridSelection).toEqual(selection);
```
ryan-bradford, line 378: Same comment as above

ryan-bradford (approved):

projects/components/src/datagrid/datagrid.component.spec.ts
irahov: yes, I have added that check also

PR #354, opened 2021-05-07
Selected entities can take a single element

ps37:
## PR Checklist

-   [X] Tests for the changes have been added (for bug fixes/features)

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
In the case of data grids with a single selection, selected entities now accept a single element as input and it does not recalculate action visibility if null is passed.

- Added a unit test to make sure that when null is passed, action availability does not get re-calculated
- Tested in VCD UI that when null is passed as input to selected entities, visibility of actions is not calculated
- The following MRS in VCD UI depends on this change. They also help in testing the above statement.
https://gitlab.eng.vmware.com/core-build/vcd_ui/-/merge_requests/3717
https://gitlab.eng.vmware.com/core-build/vcd_ui/-/merge_requests/3710

-   [X] Yes

Changed:
@Input set selectedEntities  now take a single item as a convenience to callers

Deleted:
This introduces a breaking change because `get selectedEntities` is changed to `getSelectedEntities()` as types of the setter and getter of `selectedEntities` are not the same.

### DIFF
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -8,10 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ## [2.0.0-dev.17]
 ### Changed
-- ActionMenuComponent's @Input set selectedEntities now also takes a single item as a convenience to callers
-- BREAKING: ActionMenuComponent's get selectedEntities(): T[] was replaced with getSelectedEntitites(): T[] because
-changing the getter to take T | T[] to match the setter would cause callers to do type checking even though we guarantee
-it will be an array
+- `ActionMenuComponent`'s `@Input set selectedEntities` now also takes a single item as a convenience to callers
+- BREAKING: `ActionMenuComponent`'s `get selectedEntities(): T[]` was replaced with `getSelectedEntitites(): T[]`
+because changing the getter to take `T | T[]` to match the setter would cause callers to do type checking even though
+we guarantee it will be an array.
 
 ## [2.0.0-dev.16]
 ### Added


projects/components/CHANGELOG.MD
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## [2.0.0-dev.17]
+### Changed
+- The selectedEntities input of ActionMenuComponent now takes a single item as a convenience to callers
```
juanmendes, line 11: Referring to it as `@Input set selectedEntities`  makes it clearer.

- `ActionMenuComponent`'s `@Input set selectedEntities` now also takes a single item as a convenience to callers
- BREAKING: `ActionMenuComponent`'s `get selectedEntities(): T[]` was replaced with `getSelectedEntitites(): T[]` because changing the getter to take `T | T[]` to match the setter would cause callers to do type checking even though we guarantee it will be an array.

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -249,6 +233,32 @@ export class ActionMenuComponent<R, T> {
...
+        if (this.isShallowEqual(val, this._selectedEntities)) {
+            return;
+        }
+        this._selectedEntities = val;
+        this.updateDisplayedActions();
+    }
+    private _selectedEntities: R[] = [];
+
+    /**
+     * This not a getter because getter always returns an array while the setter above can accept an array or a single
```
juanmendes, line 255: Returns the selected entities. Note: this is not a getter because its matching setter can accept an array or a single item but this method always returns an array.

projects/components/src/action-menu/action-menu.component.ts
```diff
@@ -372,7 +383,7 @@ export class ActionMenuComponent<R, T> {
     }
 
     private getContextualFeaturedActions(): ActionItemInternal<R, T>[] {
-        if (!this.selectedEntities?.length) {
+        if (!this.getSelectedEntities()?.length) {
```
juanmendes, line 386: Don't we make sure that we always return an array? Why do we to null check it?

PR #361, opened 2021-05-13
Configure quick search filters through service

ryan-bradford:
## PR Checklist

## PR Type

-   [ ] Bugfix
-   [X] Feature
-   [X] Refactoring (no functional changes, no api changes)
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?

Filters are now configured through the search service rather than through quick search itself. In addition, the search service now has methods to select a filter value and to test if the modal is pinned or not.

1. Loaded into VCD_UI
2. Set organizations page to automatically select a filter only if not pinned
3. Navigated to orgs page, observed added filter
4. Navigated away and deselected filter
5. Pinned modal
6. Navigated to orgs page
7. Observed filter NOT selected

auto-filters

examples

-   [X] Yes
-   [ ] No

We felt like filters should be configured through the service rather than through the components itself. This is like how providers are already configured. Clients should rework their code to not bind to the `@Input() filters` and instead use the `registerFilters` method.

### DIFF
--- a/projects/components/src/quick-search/quick-search.component.spec.ts
+++ b/projects/components/src/quick-search/quick-search.component.spec.ts
@@ -832,7 +832,7 @@ describe('QuickSearchComponent', () => {
         });
 
         it('adds a filter icon that shows providers only of that type', function (this: Test): void {
-            this.finder.hostComponent.filters = [
+            this.quickSearchData.spotlightSearchService.registerFilters([
                 {
                     id: 'type',
                     options: [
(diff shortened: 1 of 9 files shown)


projects/components/src/quick-search/quick-search.component.spec.ts
```diff
@@ -855,6 +855,30 @@ describe('QuickSearchComponent', () => {
             expect(this.quickSearch.getSearchResultSectionTitles().length()).toEqual(1);
             expect(this.quickSearch.getSearchResultItems().text()).toEqual('copy');
         });
+
+        it('can set filter values through the QuickSearchService', function (this: Test): void {
```
juanmendes, line 859: I don't see a need to mention quick search service as part of this test. The QuickSearch Service isn't doing much...

I do think it should be doing the filtering. We could move a lot of code to the service, simplifying the UI, its tests, and this functionality can be tested without the need for a UI.

Please create a task sharing your thoughts on how we can move all the data processing to the quick service.

juanmendes (changes requested):

projects/components/src/quick-search/quick-search.component.spec.ts
ryan-bradford: https://jira.eng.vmware.com/browse/VDUCC-694

juanmendes (approved):

PR #371, opened 2021-05-28
Removes internalFactory from WidgetObject

juanmendes:
## PR Checklist

## PR Type
-   [x] Feature

## What does this change do?

WidgetObject#internalFactory has caused confusion for developers. It caused a different way to interact with components depending on whether you're inside a `WidgetObject` or you're within a test (stepdef or unit test). 

To avoid this confusion, we now always expose `WidgetObjectElement`s and callers call `unwrap()` on their own if they want to use the `T` functionality that is specific to the testing framework being used.

Also removed the explicit return types for WidgetObject methods since they will all (mostly) return WidgetObjectElement or void for methods that abstract interactions.

Loaded the examples application, clicked through each example, verifying one piece of functionality for each

-   [x] Yes

Wherever one was using `internalFactory` to create methods to find elements, one can just use `factory` and if one needs to access native functionality, they can call `unwrap()` from its call sites.

### DIFF
--- a/projects/components/CHANGELOG.MD
+++ b/projects/components/CHANGELOG.MD
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.0.0-dev.24]
+## Changed
+- BREAKING: internalFactory has been removed from WidgetObject. All users of our exposed WidgetObjectElements
+  will need to call unwrap() on the methods returning HTML elements if they want to access `T` functionality
+
 ## [2.0.0-dev.23]
 ### Changed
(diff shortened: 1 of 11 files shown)


projects/components/CHANGELOG.MD
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Changed
+- BREAKING: internalFactory has been removed from WidgetObject
```
ryan-bradford, line 11: Should we mention that this changes the interface for all of our exported widgets?

ryan-bradford (approved):

PR #388, opened 2021-11-09
Create flat list of routes and their component's tags

juanmendes:
## What does this change do?
A new option to generate a file with a flat list of objects matching the routes that are available in the JSON output

By running
> route-analyzer -e entry-file-path --outputTs /path/to/routes.ts

A file  will be created containing  objects containing a `route` function and a `tagName` property. This can be used for:

* Automated testing that  crawls the entire application verifying that the component is rendered on the page.
* Type safe URLs, if the path changes, the object changes causing compilation errors

### Example output
```typescript 
export const libraries_vappTemplates_vAppTemplateId_vcdVappVms_vmId_general =  {
    route(vAppTemplateId: string, vmId: string) {
        return `/libraries/vapp-templates/${vAppTemplateId}/vcd-vapp-vms/${vmId}/general`;
    },
    tagName: "vcd-vms-general-details",
}
export const libraries_vappTemplates_vAppTemplateId_vcdVappVms_vmId_guestCustomization =  {
    route(vAppTemplateId: string, vmId: string) {
        return `/libraries/vapp-templates/${vAppTemplateId}/vcd-vapp-vms/${vmId}/guest-customization`;
    },
    tagName: "vcd-vms-guest-customization",
}
export const libraries_vappTemplates_vAppTemplateId_vcdVappVms_vmId_guestProperties =  {
    route(vAppTemplateId: string, vmId: string) {
        return `/libraries/vapp-templates/${vAppTemplateId}/vcd-vapp-vms/${vmId}/guest-properties`;
    },
    tagName: "vcd-vms-guest-properties",
}
```

The old  `-o` option is deprecated since there are two types of files that can be output now. See the updated README file for a description. It will be translated to the `--outputJson` and removed at a later date.

* Ran the updated script against vcd_ui's internal Angular application
* Verified generated output file (snippet in main description)
* See readme for instructions on using npm link to test changes

## Code changes
* The route analyzer now adds a `tagName` to each route, it's the `selector` for the component defined for that route.
* A new file is added which just processes the generated list of routes
* Unit tests were added for the processing of this new file
* The interfaces ModuleRouteCall,  ModuleRoutes and  ModuleAppRoutes had its module property point to the whole source file instead of just the filename. This is so we can read the module to get its `selector` that is passed to the Component Decorator
* Updated tsconfig so I could use `matchAll`

## README updates
There used to be some instructions on how to run route-analyzer without installing it from `npm`. That is unnecessary, we can always use `npm link` to achieve the same thing when developing the library and testing it from an application without publishing it.

Also, if someone want to use this library, they should be using a package manager instead of trying to run its files directly.

-   [x] No

The `--output ` option is deprecated in favor of `--outputJson` since there are two files the can be output now.

## PR Checklist

## PR Type

-   [x] Feature
-   [x] Version bump


projects/route-analyzer/src/cli.ts
```diff
@@ -17,26 +19,43 @@ const argv = yargs
     })
     .option('outputFile', {
         alias: 'o',
-        description: 'Entry file of your angular application',
+        description: 'JSON routes file to be generated by this script',
+        deprecated: 'Use outputJson instead',
+        type: 'string',
+    })
+    .option('outputJson', {
```
irahov, line 26: What about the idea that you had of specifying the output directory?
It is a little bit awkward to specify an output for file type - json & typescript

projects/route-analyzer/src/lib/route-analyzer.ts
```diff
@@ -179,15 +182,58 @@ function routeCallToRoutes(moduleRouteCall: ModuleRouteCall, typeChecker: ts.Typ
...
                 // Maybe not really needed, but just in case we add the component name
+                // What is more useful is the selector for the component
+                // Analyze the imports, looking for the component name so we can find its file
+                // And get its tag name
                 route.component = prop.initializer.getText();
+                const tagName = getTagName(prop.initializer, typeChecker);
+                if (tagName) {
+                    // Don't lose existing data that may or may not have been added, depending on the order
+                    // that properties were declared
+                    route.data = { ...(route.data || {}), tagName };
```
irahov, line 193: Isn't it better to  add a function that preserves the old data and adds the new one?
The function name will suggest that we don't lose the existing data

ryan-bradford (approved):
Looks great!

projects/route-analyzer/src/cli.ts
juanmendes: If we're OK with breaking backwards compatibility, I could do the directory, but that forces us to hard code a name. @ryan-bradford What do you think?

ryan-bradford: I think output directory might be nice. Can we just support both? Like four options:
1. Specify output directory
2. Manually specify both filed
3. Specify just JSON
4. Specify just TS

PR #426, opened 2022-04-06
Add ng live docs

juanmendes:
## What does this change do?
Move https://github.com/vmware/live-docs into this repo

One commit is [just copying files over](https://github.com/vmware/vmware-cloud-director-ui-components/pull/426/commits/3cf858a62b85037559ed987e05f7452c0b5cef52). The other commit is [what was required to get it working](https://github.com/vmware/vmware-cloud-director-ui-components/pull/426/commits/5f54c80eccc6c71f107b497fc4bf6e4939c342fd).

We never fully developed live-docs to be a multipurpose library for multiple frameworks.  Since the team has shrunk, we won't be able to do that. Having a single repo will save a lot of time when upgrading Angular for VCD.

* Ran the application and went through each of the components interacting with at least one example each.
* Created an example using the ng schematic from ng-live-docs
* Verified that [tests for ng-live-docs are being run](https://github.com/vmware/vmware-cloud-director-ui-components/runs/5856348655?check_suite_focus=true#step:6:704)

-   [x] No

## Other information
Snuck in an update of `angular-cli-ghpages` which had been [causing the deployment of the website to fail](https://github.com/vmware/vmware-cloud-director-ui-components/runs/5672436784?check_suite_focus=true).


juanmendes:
Forgot to update the readme including ng live docs information, it can just link to the existing doc in projects/ng-live-docs

package.json
```diff
@@ -51,27 +56,28 @@
...
-    "prismjs": "1.25.0",
+    "prismjs": "^1.23.0",
     "properties": "1.2.1",
-    "rbradford-compodoc": "1.1.11",
+    "rbradford-compodoc": "^1.1.11",
     "rxjs": "6.5.5",
+    "schematics-utilities": "^2.0.2",
     "systemjs": "0.20.19",
-    "tsickle": "0.39.1",
+    "tsickle": "^0.39.1",
```
ryan-bradford, line 80: NIT: exact versions?

ryan-bradford (approved):

package.json
juanmendes: Gave up using exact version, they caused too much trouble when two packages depended on the same version... The real benefit was so it wouldn't be changed when something else is installed. However, doing that may force you to have multiple version of packages causing bundle blot

PR #449, opened 2022-10-05
FormSelectComponent not taking in translation params

xvu-vmware:
## Description
The `FormSelectComponent` was hardcoded to use its control's `.value` to be passed to the translation service (along with its key) to display the error messages.

Now, instead of passing its `.value`, it will use the values provided by the `ValidationErrors`'s values. The translation parameters could be `any` or an array of `any`. This get pass to the TranslationService to be used for translation.

## Example

```typescript
// My custom validator
function isAllGoodValidator(control: FormControl) {
   if (control.value != "SaulGoodman") {
      return {'some.error.key': [ 'all', 'is', 'man', 'good', 'It']}
   }
   return null;
}

// In en.properties
// some.error.key={4} {1} {0} {3}, {2}!
```

The select above would show the following error message: `It is all good, man!`
   

Signed-off-by: Xuanvinh Vu xvu@vmware.com

## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
Before this change, the FormSelectComponent translates its validation keys with the selected option's value. Sometime this is not the behavior we want so with this change the user can pass in the translation parameters to be use with the translation key(s). To do this a new method called `translatedErrors()` is introduced in the `BaseFormControl`. When a form input component wants to get the translated error messages all it needs is to call `translatedErrors()`.

We should refactor other form input components to use this new method instead of translating the error messages themselves. https://github.com/vmware/vmware-cloud-director-ui-components/issues/450

 Manually tested the "Select form input validation" example.
- Select the empty option from the select input and verify that validation messages are shown. The messages are "Required" and "Only numbers in the range of one to three are allowed".
- Select value from "one" to "three" and verify that there are no validation message shown.
- Select "four" from the select input and verify that validation message "Only numbers in the range of one to three are allowed" is shown.

Validator that return key/value where the value is an array to be passed to translation service
image

Multiple validators
image

-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/components/src/form/base-form-control.ts
+++ b/projects/components/src/form/base-form-control.ts
@@ -5,6 +5,7 @@
 
 import { Directive, Input } from '@angular/core';
 import { AbstractControl, ControlValueAccessor, FormControl, NgControl, ValidatorFn } from '@angular/forms';
+import { TranslationService } from '@vcd/i18n';
 import { IdGenerator } from '../utils/id-generator/id-generator';
 import { CanBeReadOnly } from './interfaces/can-be-read-only.interface';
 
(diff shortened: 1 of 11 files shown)


xvu-vmware:
Updated pull request with select form input validation example.

image

PR #482, opened 2023-01-17
Fix route-analyzer not handling equal module basenames

kepelrs:
Before this commit, the route-analyzer would search for modules by their basenames (eg. my-module.ts). This could match the wrong module in cases where we had two module files with same file name but in different folders.

This change fixes the issue by matching absolute paths instead of only basenames.

Signed-off-by: Davi Barreto <dbarreto@vmware.com>

## PR Checklist

## PR Type

-   [x] Bugfix

## What does this change do?
This change makes the route-analyzer correctly handle duplicate file names by using the absolute path.

### Test new output is unchanged when there are no duplicate module names
    a. Run the old route-analyzer locally against VCD master
    b. Store the contents of `gen-routes/tenant/app-route.ts` into a separate file
    c. Run the newly changed route-analyzer locally against VCD
    d. Verify the generated `app-route.ts` file from step a. is strictly equal to step c.
### Test the old script breaks when duplicate modules are found
    a. checkout vcd_ui master
    b. cd content/core
    c. Run the following commands to setup a duplicate module with loadChildren:
    d.  ng g m z-module --routing
    e.  ng g m z-module/catalogs --routing --route catalogs --module=z-module
    f. change app.route.tenant.ts line 32, make it point to ZModuleModule
    g.  npm run gen-app-routes
    h. verify an error is thrown
### Test the new script works when duplicate modules are found
    a. Perform steps a. through f. from the previous test case
    b. Place the fixed route-analyzer under the node_modules folder
    c. npm run gen-app-routes
    d. verify no error is thrown

    
-   [ ] Yes
-   [x] No

### DIFF
--- a/projects/route-analyzer/CHANGELOG.MD
+++ b/projects/route-analyzer/CHANGELOG.MD
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### [Fixed]
+- Fixed bug that broke the route-analyzer when two routing modules had the same base name
 
 ## [0.0.6]
 ### [Added]
--- a/projects/route-analyzer/src/lib/route-analyzer.ts
+++ b/projects/route-analyzer/src/lib/route-analyzer.ts
@@ -2,6 +2,7 @@
  * Copyright 2020 VMware, Inc.
  * SPDX-License-Identifier: BSD-2-Clause
  */
+import * as path from 'path';
 import * as ts from 'typescript';
 import { AppRoute } from './app-route';
 import { ANGULAR_CORE, ROUTE_PROP } from './constants';
@@ -54,7 +55,11 @@ export function getRoutes(program: ts.Program): AppRoute[] {
         .filter(hasValue) // Not all the files that contain router usage, contain router configuration call
         .map((moduleRouteCall) => routeCallToRoutes(moduleRouteCall, program.getTypeChecker()));
 
-    const moduleAppRoutes: ModuleAppRoutes[] = fixupLoadChildren(moduleRoutes, moduleFiles);
+    const moduleAppRoutes: ModuleAppRoutes[] = fixupLoadChildren(
+        moduleRoutes,
+        moduleFiles,
+        program.getCurrentDirectory()
+    );
 
     // Get only the routes into a single array
     const appRoutes: AppRoute[] = moduleAppRoutes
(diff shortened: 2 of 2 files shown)


projects/route-analyzer/src/lib/route-analyzer.ts
```diff
@@ -301,4 +307,22 @@ function fixupLoadChildren(
...
+        let absolutePath: string;
+
+        // root filenames are user provided and may be relative (eg. see package json's gen-app-routes:tenant)
+        if (path.isAbsolute(moduleFolderPath)) {
+            absolutePath = path.join(moduleFolderPath, loadChildrenImportPath);
+        } else {
+            absolutePath = path.join(programDirectory, moduleFolderPath, loadChildrenImportPath);
+        }
+
+        // '.ts' is often omitted on TS imports.
```
juanmendes, line 322: Isn't the ts extension disallowed in imports? The compiler is often looking for `.d.ts` files or it could even be a JS file? Can you explain in a comment when `.ts` is present?

kepelrs: I got it mixed up with js. You are correct, this never happens... And never will happen. Found this read to be interesting: https://github.com/microsoft/TypeScript/issues/35589

projects/route-analyzer/src/lib/route-analyzer.ts
```diff
@@ -301,4 +307,20 @@ function fixupLoadChildren(
         }
         return referencedModule.module;
     }
+
+    // Using absolute path prevents issues when two module files have the exact same name
+    function getLoadChildrenAbsolutePath(moduleFolderPath: string, loadChildrenImportPath: string): string {
+        let absolutePath: string;
```
juanmendes, line 313: Nitpicking since `const` is preferred

```ts
const absolutePath = path.isAbsolute(moduleFolderPath 
   ?  path.join(moduleFolderPath, loadChildrenImportPath) 
   :  path.join(programDirectory, moduleFolderPath, loadChildrenImportPath);

// Must append the file extension, as the TS imports never include it
return `${absolutePath}.ts`;
```

projects/route-analyzer/CHANGELOG.MD
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Fix route-analyzer to use absolute paths instead of basenames
```
juanmendes, line 9: Please add the header  and include something that helps the user know what happened instead of a description of the code change.


```
### [Fixed]
Fixed bug that broke the route-analyzer when two routing modules had the same base name
```


`### Fixed`

juanmendes (approved):

juanmendes (approved):

PR #532, opened 2023-08-24
Introduce a new `@Input() ariaLabel` to be used when input's label is not displayed.

ttsvetko:
Introduce a new `@Input() ariaLabel` param for `vcd-form-input` components. The string input is used to set aria label when label is not provided and displayed.

Description
===========
input elements should be associated with label or should have aria-label attribute set.

Fix Description
===============
introduce a new `@Input() ariaLabel` param for `vcd-form-input` components. The string input is used to set aria label when label is not provided and displayed.

## PR Checklist

## PR Type

-   [x] Bugfix
-   [ ] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [ ] Version bump

## What does this change do?
Introduce a new `@Input() ariaLabel` param for `vcd-form-input` components. The string input is used to set aria label when label is not provided and displayed.

* Have a vcd-form-input with `@Input() label` set
  * the input is associated with <label> element
* Have a vcd-form-input with `@Input ariaLabel` set and no `@Input label` set
  * the input has an `aria-label` attribute set
* Have a vcd-form-input with both `@Input() label` and `@Input ariaLabel` set
  * the input should be associated with <label> element and no aria-label attribute should be set for the input.

-   [ ] Yes
-   [ ] No


projects/components/src/form/base-form-control.ts
```diff
@@ -102,6 +102,11 @@ export class BaseFormControl<V> implements ControlValueAccessor, CanBeReadOnly {
      */
     @Input() isResponsive = true;
 
+    /**
+     * Aria label to be set when label input is not set
+     */
+    @Input() ariaLabel: string;
```
juanmendes, line 108: We must provide either `ariaLabel` or `label`. That's a pattern that's good to avoid. Initially I was going to ask that you document the `label` to know that you must provide the `ariaLabel` if you don't provide this. Maybe we should error if we detect this error condition in `ngOnInit` . Then I started trying to think of a way to avoid the problem altogether.

if instead, we added a new `@Input() hideLabel=false`? Then it just uses the label input as the `aria-label`? It makes sure someone doesn't forget to set either and shields the user from needing to understand about aria.

juanmendes: Angular 16 is going to allow for required inputs, and label should be required. It can currently be hacked to be required. See https://indepth.dev/posts/1517/required-inputs-in-angular-v16

ttsvetko: I also had the same thoughts considering something like `hideLabel`, but I did not like that my idea because I have a `label` input and I would expect it won't be displayed if I pass empty string or just do not pass it. And that `hideLabel` sounds strange when users can just not pass the label. That's why I ended up with that aria-label. I also thought that we may have some eslint rule that will check that ariaLabel is set when no label is passed - this will ensure all form controls will meet the a11y requirement. 

Now that you are thinking also in that direction makes me think that this could be the better approach. In addition to that we may make the label as required param when we upgrade to angular 16. But it still looks odd to me that we are asking for a label and having hideLabel=true.

Let me know if you insist in the `hideLabel` input and I will update the code.

juanmendes:
The commits should not start with

"If this this commit is added it will:" That is just guidance on the format. This makes most of the commit message not readable when looking at commit lists. Write the full sentence but don't leave in the guidance. Just like you did to the MR title.

The same guidance text should be replaced in "testing done"

"First use case" is not descriptive, it's just a guide. You should describe your first use case (WIth label) and your second case is (Without label) and all the different use cases you have tested.

projects/components/src/form/base-form-control.ts
juanmendes: I like it because:
*  We can enforce that a label must always provided. We can enforce it even before v16 as https://indepth.dev/posts/1517/required-inputs-in-angular-v16  shows.
* The user doesn't need to understand about aria to get a11y goodness. The code will guide them that they must pass in a label. The documentation can easily state that if you don't want it to be displayed, you can set it to hidden using `hideLabel`. Maybe we don't even need a new `hideLabel` we can just make label be of type `string |  HiddenLabel`.


Feel free to ask for feedback from others.

ttsvetko: As we agreed in our internal Slack discussion and poll - we will have a new `@Input hideLabel` that will be used for hiding the label. We will assume the `label` value is always set until we migrate to Angular 16 and be able to make it required.
Changes have been applied and this merge request could be reviewed again.

PR #537, opened 2023-10-05
Fix clr-datagrid matching the wrong trackBy for selection comparison

kepelrs:
Clarity introduced a new 'clrDgItemsTrackBy' input property which is used to compare the identity of selected grid items.

This new property has a different signature than the regular trackBy function, which is why we are taking our client's provided trackBy and wrapping it before passing into clrDgItemsTrackBy.

Link for the original bug: https://github.com/vmware-archive/clarity/issues/3936
Link for clarity's fix: https://github.com/vmware-clarity/ng-clarity/pull/539

## PR Checklist

## PR Type

-   [x] Bugfix

## What does this change do?
Fixes clarity's ["Datagrid uses wrong trackBy and breaks selection"](https://github.com/vmware-archive/clarity/issues/3936) bug by applying their [bugfix](https://github.com/vmware-clarity/ng-clarity/pull/539).

Tested against VCD's codebase.

-   [x] No

### DIFF
--- a/projects/components/src/datagrid/datagrid.component.ts
+++ b/projects/components/src/datagrid/datagrid.component.ts
@@ -375,7 +375,9 @@ export class DatagridComponent<R extends B, B = any> implements OnInit, AfterVie
         private translationService: TranslationService,
         private changeDetectorRef: ChangeDetectorRef,
         private subTracker: SubscriptionTracker
-    ) {}
+    ) {
+        this.trackBy = (index: number, record): string => (record as any).href;
+    }
 
     /**
      * The pagination information that the user should supply.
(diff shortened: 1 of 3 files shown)


projects/components/src/datagrid/datagrid.component.ts
```diff
@@ -735,9 +740,14 @@ export class DatagridComponent<R extends B, B = any> implements OnInit, AfterVie
      *
      * If the record has a href, defaults to that. Else, defaults to index.
      */
-    @Input() trackBy: TrackByFunction<B> = (index: number, record): string => {
-        return (record as any).href;
-    };
+    @Input() set trackBy(trackByFn: TrackByFunction<B>) {
+        this._trackBy = trackByFn;
+        this.clrDgItemsTrackBy = (record) => trackByFn(undefined, record);
```
kepelrs, line 745: I'm considering if we should pass a bogus `0` instead of `undefined`, to ensure we are not introducing runtime errors in the off chance someone is doing something with the `index: number`. Thoughts?

juanmendes (approved):

projects/components/src/datagrid/datagrid.component.ts
kepelrs: Replaced `undefined` with `0`, filed a clarity issue and included a comment with a link to it.

juanmendes (approved):

PR #542, opened 2023-11-27
Datagrid filter setFocus

irahov:
## PR Checklist

## PR Type

-   [x] Bugfix
-   [x] Feature
-   [ ] Build related changes
-   [ ] CI related changes
-   [x] Version bump

## What does this change do?
When opening a datagrid filter the focus should go directly to the relevant HTML element so that the user can start filtering right away.
Up to now the focus was on the flter close button 'X'. This is extremely boring when using string filter because the user has to click first on the input field (or TAB to it) and then start writing.

In the example applications navigate to `Datagrid` -> `Datagrid Filters` (`/datagrid/example/datagrid-filter`)
Open each of the filters and verify that the focus is on the `input` field or the `select` .

-   [x] Yes
-   [ ] No

Breaking changes need:
- removing  modifier for ClrDatagridFilter parameter of the child constructor, i.e. it should be `filterContainer: ClrDatagridFilter` only
- implement setFocus() method on the child class

### DIFF
--- a/projects/components/src/datagrid/filters/datagrid-filter.ts
+++ b/projects/components/src/datagrid/filters/datagrid-filter.ts
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2019 VMware, Inc.
+ * Copyright 2019-2023 VMware, Inc.
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
@@ -64,7 +64,10 @@ export abstract class DatagridFilter<V, C extends FilterConfig<V>, F extends For
 {
     abstract formGroup: F;
 
-    protected constructor(filterContainer: ClrDatagridFilter, private subscriptionTracker: SubscriptionTracker) {
+    protected constructor(
+        protected filterContainer: ClrDatagridFilter,
+        private subscriptionTracker: SubscriptionTracker
+    ) {
         filterContainer.setFilter(this);
     }
 
--- a/projects/components/src/datagrid/filters/datagrid-filter.ts
+++ b/projects/components/src/datagrid/filters/datagrid-filter.ts
@@ -6,7 +6,7 @@
 import { Directive, Input, OnInit } from '@angular/core';
 import { FormGroup } from '@angular/forms';
 import { ClrDatagridFilter, ClrDatagridFilterInterface } from '@clr/angular';
-import { Subject } from 'rxjs';
+import { Subject, timer } from 'rxjs';
 import { debounceTime } from 'rxjs/operators';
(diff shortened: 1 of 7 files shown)


projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
```diff
@@ -127,6 +127,10 @@ export class DatagridMultiSelectFilterComponent extends DatagridFilter<string[],
             !!Object.keys(this.formGroup.getRawValue()).filter((frmCtrl) => this.formGroup.get(frmCtrl).value).length
         );
     }
+
+    protected setFocus() {
```
juanmendes, line 131: I think a better approach is to have a method called `getFocusableElement()`, then the callers just have to return the element instead of every implementation calling `focus()`.

projects/components/src/datagrid/filters/datagrid-filter.ts
```diff
@@ -64,7 +64,10 @@ export abstract class DatagridFilter<V, C extends FilterConfig<V>, F extends For
 {
     abstract formGroup: F;
 
-    protected constructor(filterContainer: ClrDatagridFilter, private subscriptionTracker: SubscriptionTracker) {
+    protected constructor(
+        protected filterContainer: ClrDatagridFilter,
+        private subscriptionTracker: SubscriptionTracker
```
juanmendes, line 69: Since we're touching this file, we may as well remove the providers from here since it doesn't have any effect, the concrete classes need to add the providers.

juanmendes: @kepelrs discovered this while analyzing the SubscriptionTracker

projects/components/src/datagrid/filters/datagrid-multiselect-filter.component.ts
irahov: I like the `focus()` since it seems a more complete approach to me.
If I come accross to `getFocusableElement()` function - it is not that clear how the returned element will be used. Yes, the name suggests to focus, but yet - not that clean. In addition, I'm not sure if there is a case when something else may be needed when focusing the element. On the other hand `focus()` is more concise and clear.

I agree that the very `.focus()` call can be reused in the base class but I don't see it has the advantages as the other approach.

projects/components/src/datagrid/filters/datagrid-filter.ts
kepelrs: IIRC it was @irahov who discovered it first :)

PR #544, opened 2023-12-04
Error in activity reporter not visible, need to scroll to the top

Juliana216:
Bugfix: Add a scrollToMessage method for the vcd-banner-activity-reporter. This increases the visibility of success and error messages by having them in the view of the user.

Testing Done:
- Edited examples locally to have a page where the error/success message is not visible.
- Observed that the scrolling to the error/success message worked.

https://github.com/vmware/vmware-cloud-director-ui-components/assets/40737169/4950a5df-1a69-427c-b516-3cd2cc22bb9f

[COR-12499](https://jira.eng.vmware.com/browse/COR-12499)
https://bugzilla.eng.vmware.com/show_bug.cgi?id=2858757

### DIFF
--- a/projects/components/src/common/activity-reporter/banner-activity-reporter.component.ts
+++ b/projects/components/src/common/activity-reporter/banner-activity-reporter.component.ts
@@ -1,9 +1,9 @@
 /*!
- * Copyright 2020 VMware, Inc.
+ * Copyright 2020-2023 VMware, Inc.
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
-import { Component, Inject, Input } from '@angular/core';
+import { Component, ElementRef, Inject, Input } from '@angular/core';
 import { TranslationService } from '@vcd/i18n';
 import { LazyString } from '@vcd/i18n';
 import { ActivityPromiseResolver } from './activity-promise-resolver';
@@ -29,6 +29,7 @@ export class BannerActivityReporterComponent extends ActivityReporter {
 
     constructor(
         private translationService: TranslationService,
+        private elementRef: ElementRef,
         @Inject(ActivityPromiseResolver) promiseResolver: ActivityPromiseResolver<any>
     ) {
         super(promiseResolver);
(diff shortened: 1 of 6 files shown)


projects/components/src/common/activity-reporter/banner-activity-reporter.component.ts
```diff
@@ -75,6 +77,14 @@ export class BannerActivityReporterComponent extends ActivityReporter {
         this.successMessage = null;
     }
 
+    /**
+     * To view error and success messages, scroll to the top of the page for visibility
+     */
+    scrollToMessage(): void {
+        const element = document.querySelector('vcd-banner-activity-reporter');
```
juanmendes: This is not a safe way, you could have two banners on the page. The correct way is to get a reference to the component's element and query using that.

If you are ever forced to use `document` in Angular, you should always think twice, and if really required, document why.

This method should be private since we should always minimize a component's public API. 

The method name is misleading since it's not scrolling the component itself, it's scrolling components around it so that it comes into view, please rename it like the HTML method, `scrollIntoView`. The comment is also either inaccurate or ambiguous since you are not necessarily scrolling to the top of the page. Scrolling to the top of the page may not bring an element into view, it may need to be scrolled down so that it's visible, don't embed the assumption that this component must be used at the top of a page.

The variable that was created doesn't help code comprehension. it's only used once after it was created and its name is very generic.

```ts
constructor(private elementRef: ElementRef) { }

private scrollIntoView: void {
  this.elementRef.nativeElement.scrollIntoView();
}
```

juanmendes:
Please create a new example  [under activity reporter ](https://vmware.github.io/vmware-cloud-director-ui-components/bannerActivityReporter/example)that shows this behavior

juanmendes (changes requested):

projects/components/src/common/activity-reporter/banner-activity-reporter.component.ts
Juliana216: Addressed the comments from above in my new changes.

